PR #6, opened 2016-04-15
update npm install name in README

kurttheviking:
Minor fix to prompt folks to use the org-managed package from npm


robertleeplummerjr:
Nice!

PR #18, opened 2016-07-07
Closes #17

### LINKED ISSUE #17: get browser version running
Would also be nice while doing this to consider where we absolutely need third party plugins.  They are used occasionally, but nothing seemingly to complex for native js to handle.

robertleeplummerjr:
browser back online with updated copywrite
remove dependency on underscore
rename neuralnetwork to neural-network to be consistent
put train-stream in its own file
bump version number


robertleeplummerjr:
I have a pet peeve.  I feel like, while what browserify is doing is noble, there really isn't a true need for it here.  A careful glance over the lib and I find with my imperfect eye the only thing (now in this pr) that cannot run in the browser is `TrainStream`.  Of course, I'm open to suggestions, we could easily concatenate all the files, stripping `var it = require(...` and we'd have a 0 dependency browser version of brain.js.

Thoughts?

robertleeplummerjr:
#17 for linking

robertleeplummerjr:
Let's say we just concatenate everything together minus TrainStream, as a string, then we could do this:

``` js
var browserVersion =
  'var brain = (function() {'
    + source
    /** removes
      * var name = require('name');
      * var aName = require('a-name');
      * var name = require('name'),
      *  aName = require('a-name');
      * */
      .replace(/(\n*(var)?\s*[A-Za-z]+\s*[=]\s*require[(]['][a-z-/]+['][)][;,][\n]*)+/g, '')

    /** removes
      * module.exports = Something;
      * */
      .replace(/\n*module[.]exports\s*[=]\s*[A-Za-z]+[;]/g, '')
    + '; return brain;})()';
```

and bam!  Browser version.

IonicaBizau:
To make my npm packages compatible with the browser environment I use [`dist-it`](https://github.com/IonicaBizau/dist-it) which uses browserify.

Actually, what is the limitation here? Why isn't `TrainStream` working in the browser?

robertleeplummerjr:
https://github.com/harthur-org/brain.js/pull/18/files#diff-2172854117591ffdafbb876861403401R2 it uses  node's `stream`

IonicaBizau:
@robertleeplummerjr But `browserify` will probably take care of that too, right?

robertleeplummerjr:
touche` sir, touche`

robertleeplummerjr:
I added both browserify and uglify to grunt's default command, along with unit tests, so now we have one easy means of both testing, and deploying.  Should be ready for review now.

PR #19, opened 2016-07-22
Another example to README

AlibekJ:
This example illustrates that input objects can have different structure.


robertleeplummerjr:
landed

robertleeplummerjr:
Also, thank you for your fine addition!

PR #28, opened 2016-09-06
es6, here we come

robertleeplummerjr:
bam

it just got cooler


robertleeplummerjr:
Since there is no logic changes, other than bug fixes, with the unit tests to back it, and because this is such a widely wanted feature...  Merging.

Issue #11, opened 2016-05-22
es6 it!


robertleeplummerjr:
I'd like to update the syntax to that of es6, thoughts?

MrIsaacs:
Seems legit. Maybe having an own branch for es6 is a good thing? So tagging different branches is made easier an makes the codebase compatible for older interpreter and users can choose a compatible version as they need.

robertleeplummerjr:
If we were to do it, I'd first create a new branch, then I'd like to ensure we use something like babel to compile it for compatibility.  I'd also like to make part of the build process create a browser specific file, so that rather than doing it manually, it is built in.  All of which wouldn't be too complicated.

robertleeplummerjr:
Once all that clears, then we could merge it into master, and continue on our merry way.

robertleeplummerjr:
It is here, huzzah!

IonicaBizau:
There's a tool caled [`lebab`](https://github.com/mohebifar/lebab) by @mohebifar which would be probably useful to convert it automagically and then edit the things that still have to be edited manually (if any). :grin:

robertleeplummerjr:
very cool!

robertleeplummerjr:
What would be our strategy for es5/legacy compatibility?

Macil:
Why not transpile it to es5 with babel as a prepublish step? That seems to be the normal route for modules written in es6+.

robertleeplummerjr:
I like it!

robertleeplummerjr:
@IonicaBizau I mostly manually translated it already.

robertleeplummerjr:
Ready and waiting here: https://github.com/harthur-org/brain.js/pull/28

robertleeplummerjr:
As soon as this is merged, auto builds are next, then a pr for the recurrent neural network stuff.  Just fyi.

robertleeplummerjr:
Merged!

IonicaBizau:
For publishing I use `babel-it` which babelifies the files in place when `npm publish`ing.

robertleeplummerjr:
Would you be up for taking a look at master, and seeing if you can apply your knowledge there?

IonicaBizau:
@robertleeplummerjr Well, pretty much `npm i -g babel-it` and instead of doing `npm publish` you do `babel-it`. 😁

robertleeplummerjr:
[...]

Issue #10, opened 2016-05-10
Demo site url /brain -> /brain.js


ecwyne:
http://harthur-org.github.io/brain.js/

craigsheen:
https://harthur.github.io/brain/ Would also work (as further down in the documentation.)

robertleeplummerjr:
Thanks for finding!
https://github.com/harthur-org/brain.js/commit/a52d12e3b5cd9e4ccce118a63e526cab40735ea4#diff-04c6e90faac2675aa89e2176d2eec7d8
https://github.com/harthur-org/brain.js/commit/55371e1be1b8fabd9d351f1bb1d7dfa3d9f0e77d#diff-04c6e90faac2675aa89e2176d2eec7d8
https://github.com/harthur-org/brain.js/commit/18e52bd6b315a21d9437dfac906e0056173dd888#diff-04c6e90faac2675aa89e2176d2eec7d8

Issue #9, opened 2016-04-19
Cross-validation tests failing


pejrak:
fails on multiple accounts:
`> mocha test/cross-validation --timeout 10000

(node) child_process: options.customFds option is deprecated. Use options.stdio instead.

  OCR cross-validation
Cross validating
    1) recognize characters in different fonts

  0 passing (563ms)
  1 failing

  1) OCR cross-validation recognize characters in different fonts:
     TypeError: Cannot read property 'learningRate' of undefined
      at testPartition (/var/www/brain.js/lib/cross-validate.js:21:28)
      at /var/www/brain.js/lib/cross-validate.js:60:18
      at Array.map (native)
      at crossValidate (/var/www/brain.js/lib/cross-validate.js:55:28)
      at Context.<anonymous> (/var/www/brain.js/test/cross-validation/ocr.js:63:18)
      at callFn (/var/www/brain.js/node_modules/mocha/lib/runnable.js:250:21)
      at Test.Runnable.run (/var/www/brain.js/node_modules/mocha/lib/runnable.js:243:7)
      at Runner.runTest (/var/www/brain.js/node_modules/mocha/lib/runner.js:373:10)
      at /var/www/brain.js/node_modules/mocha/lib/runner.js:451:12
      at next (/var/www/brain.js/node_modules/mocha/lib/runner.js:298:14)
      at /var/www/brain.js/node_modules/mocha/lib/runner.js:308:7
      at next (/var/www/brain.js/node_modules/mocha/lib/runner.js:246:23)
      at Immediate._onImmediate (/var/www/brain.js/node_modules/mocha/lib/runner.js:275:5)
      at tryOnImmediate (timers.js:534:15)
      at processImmediate [as _immediateCallback](timers.js:514:5)
`
- when added trainOpts, this is the output err:

`Training iterations per second: 45.87900933820544
    1) recognize characters in different fonts

  0 passing (3s)
  1 failing

  1) OCR cross-validation recognize characters in different fonts:
     ReferenceError: assert is not defined
      at Context.<anonymous> (/var/www/brain.js/test/cross-validation/ocr.js:84:5)
      at callFn (/var/www/brain.js/node_modules/mocha/lib/runnable.js:250:21)
      at Test.Runnable.run (/var/www/brain.js/node_modules/mocha/lib/runnable.js:243:7)
      at Runner.runTest (/var/www/brain.js/node_modules/mocha/lib/runner.js:373:10)
[...]

PR #8, opened 2016-04-19
Dependency version bumps fixing issue #7 and #9

pejrak:
Increased versions on canvas and other dependencies to allow for npm installation.

PR #33, opened 2016-11-02
changed flood function in stream-example, for loop to '<=' from '<' s…

Kalevera:
updated flood function in stream-example.js so that it will send a null chunk to trainstream.write(). This triggers finish event emitter in write. Original caused a nodejs stream3 error because nodejs object mode true doesn't allow for null data writing. 

Example would throw an error, but the change allows the for loop to send undefined to trainsteam.write, causing the chunk to be false triggering the finished event emitter.

### DIFF
--- a/examples/stream-example.js
+++ b/examples/stream-example.js
@@ -33,9 +33,7 @@ var trainStream = net.createTrainStream({
 flood(trainStream, xor);
 
 function flood(stream, data) {
-  for (var i = 0; i < data.length; i++) {
+  for (var i = 0; i <= data.length; i++) {
     stream.write(data[i]);
   }
-  // let it know we've reached the end of the data
-  stream.write(null);
 }


robertleeplummerjr:
Does this fix the behaviour of streams?

Kalevera:
This doesn't change the behavior of the current brianjs stream implementation. What it does is show how to use the current trainstream. The way trainstream was written It will emit a finish event when there's no more data. I notice that it's manually emitting 'finish', as opposed to calling write.end() which then triggers a 'finish' event. 

I suppose it was originally done this way since the use case is not trying to append anything to the end of the stream. If you think it's necessary I could see about updating the train stream to use the write.end() methodology.

robertleeplummerjr:
Whichever is the recommended use, I'm fine with.

Kalevera:
I think using the end() method would be best as its seems to be the nodejs recommended way. I'll see if I can make the change tonight without effecting anything else in the trainstream implementation. If successful I'll submit a new pull request. If not I'll report back the issue.

robertleeplummerjr:
👍

Kalevera:
managed to run a few test during lunch and just now. I now see why the finish event is being emitted. as well as why the write.end() wasn't used. 

Writable streams don't emit an 'end' event. Calling write.end() closes the stream, the way trainstream was written it's handling the write.end() event in the floodcallback() function. So to be able to trigger the event at the correct circumstance it needed to be emitted. 

There's no need to change the trainstream implementation people just need to be aware that sending undefined is being handled in the floodcallback which will re iterate training if number iterations or error conditions haven't been met. This should be good to go and the example functions as intended with the purposed change.

robertleeplummerjr:
A documentation change/addition then?

Kalevera:
What about a gotchas section? The stream-example bug is more of a gotcha if someone codes how the example is setup. The trainstream isn't broken it's just handling a false value for you, so adding '<=' is crucial to getting the stream to behave as intended. 

Also I've been messing around with different training techniques, basically training the net with different data formats and lengths. I noticed that inputs need to be larger or equal to ouput lengths when dealing with Arrays. But output lengths did not need to be the same length. I also observed some other gotchas like having the length of the first output being one and the rest being length two will cause an output of length one when testing data on a trained network. This is interesting if you're testing data like language or audio files, etc. Since the data length won't be the exact same.  I'm still playing around with the inputs and outputs. But I still haven't tested my solution, which is padding inputs and expected outputs so that they're the same length as the maximum input, and  outputs are the same length as the maximum output. 

Have you experienced this at all? It came up because I was getting NaN when tests data inputs of different length and having outputs of different lengths. Also you can't feed an array of input and receive an object as an output. And vice versa. So maybe an expansion on the docs which describes how the data may need to be formatted to get something meaningful. I imagine padding the output could cause issues with  training the net becuase the padding would mess up the weights and bias. Padding the input I don't see much of an issue. I could work on a gotcha section that covers the training data formatting issues and the stream-example issue.

robertleeplummerjr:
I have not experienced that, but a gotchas section would be fantastic.

robertleeplummerjr:
Any further input here?

Kalevera:
Yes, I've been working on training the network to do a form of language classification. I know there's "natural" for node the language classification. But they store stop words in an array, as well as store the key phrases in the array.

 I figured brainJS can do the same thing it's a matter of how to do it. 

Initially I started off by making objects with arbitrary weights that I was assigning in the input that seemed cumbersome if a person needed to build a data set. Or the user would have to build an API that weights the words for them, I figured that may not work since the words importance is based on the context. 

So I moved to reading the data with a stream the stream returns the data in a buffer in numbers 0-255. 

My approach was then have a data set that has a sentence input(string) and then a few key words related to the sentence e.g. is the sentence a question, declaration, or a request, is it about time, location, and what significant words were used.)

I was somewhat successful in doing this an this is where the gotchas come into play. 

Gotcha #1 if you build an index file and run it in node you have to wait for the training session then see what it gives you if you put in a few sentences at the end to check/ run(). I wanted to test a bunch of sentence to see where the classification was working and if changing the sentence structure would yield different results. 

Gotcha #2 
I decided to build a small javascript object that read the data from a stream, formatted the strings into an array of numbers (divided by 255, because the streams maximum is 255) so they'd be between 0 and 1.

Gotcha #3 arrays that vary significantly in length will cause brainjs to return NAN on the output. This goes for different length inputs (I noticed this when the inputs were a difference of 5 in length. And the outputs difference were 3 in length ). So I fluffed the data by appending 0.00 to the end of arrays that were short up to the maximum input length and the same for the output. 

Gotcha #4 if the decimal on a number is 0.005 ± then the letter it generates after being converted will be off by 1 to 3 indexes (eg. "q" may end up as "n"). So part of the javascript object was dedicated to holding onto the network and sending it test strings to see which words it recognized and could classify.

Gotcha #5 is about network initialization states and dataset.

    Test #1 
To get the network close to classifying the sentence correctly you need a large dataset. I start with 80 inputs and outputs, with 20,000 iterations, three layers [150,88,200], and a threshold of 0.005 or 99.5% accuracy. 

This performed poorly but would meet the threshold before the iterations completed. The issue was it could classify questions if the word "what" was present in the sentence but it didn't recognize other questions words like who, , when, where, and how.

Test #2 
I figured it's a complex task I'm asking of the network it may need more nodes and layers.

So I left the data set the same and upped the layers and nodes to see if additional abstractions on the networks Mart could figure it out. [250,380,180,222] unfortunately this caused the network to not meet the threshold and go over the iterations. Only reaching 0.138 error. Which didn't classify anything correctly.

Test #3 
Instead of adding more layers and more iterations I upped the dataset. To 100 in 100 out (now these are different classifications within the dataset not just one type like question) 

This resulted in a smaller error 0.108. this sort of classified questions but the answer was un-readable becuase the letters were so far off of that they should be. 

I thought surely I'm onto something though because the error decreased with a larger data set.

Test #4 
I then upped the dataset to 250 in and 250 out. I left the initialization the same as the prior test. The larger dataset and large node size cause these test to run around an hour maybe a little more.

This yielded an error threshold of 0.0138, which had the same problems as the above test the answers generated were un-readable since the letters were so far off. Still more of the answered yielded were visible to a human as "questions" which was represented onscreen as "pueshhsen" or "rwd'tsm$" there multiple types of messed up answers but these two were close.

Test #5
I thought well surely the dataset caused the answers to get close maybe I should up the nodes and layers again. [408,300,280,370,403,238,321] these are arbitrary choices. I left the datsset as 250in and 250 out.

This resulted in the error threshold increasing back to approximately 0.136 which I thought was coincidental to the earlier error when I increased the nodes and layers. 

This cause the training session to last 6-7 and you could see in the end of the training that it was hung up trying to decrease the threshold only making 0.0000002 increments of progress for every 100 iterations.

Test #6 
By this point I figured the issue was the test data and it needed to be much greater (image recognition for human written numbers had datasets in the thousands...) So embarked on training more data. I did 300 in and 300out. This lowered the error to 0.08 increased the testing time to about 8 hours

Test #7
Then again I upped it to 400 in and 400out

This yielded an error threshold of 0.0108. And took 12 hours (not sure why). But it maxed out the iterations.

Test #8
Upped it to 508 in and 508 out ( happen to come up with 8more sentences...) I then decreased the nodes and left the layers the same.
[208,100,250,170,203,238,121] . This caused the error to increase back to 0.185 and again was in un-acceptable. Also this maxed out the iterations counts.

This trained for about a day and a half 36 hours.

Test #9 
I went back to increased nodes and left the dataset the same and went with [408,300,280,370,403,238,321]  nodes and layers. 

The increased nodes and the increased dataset was maxing out the iterations and heck I figured this was an issue so I bumped the iterations to 30,000.

This allowed the error threshold to get back to 0.0176 but it maxed out the iterations this took 48hours to finish training.

Test #10 
Left everything the same and bumped iterations again to 40,000 and viola I made it to 0.00499 error threshold. 

This is when I found out that the 0.005 differe in a decimal will cause the letters to be off by one.  It was closer to classifying and could determine if things were a question or a declaration. They weren't spelled correctly at all but you could see that the letters were only off by one or two indexes.

Test #11
I bumped the iterations to 50,000 and the threshold to 0.0005. 

This maxed out the iterations but yielded an error threshold of 0.00208, this classified all the question words and some of the special words like color, time, you, and my. It classified most all the sleep and alarm classifications like alarm, bed and timer. Again the error wasn't small enough so almost every classification had a single letter some times two that were off by one index.

This also last 56 hours for the training session.

Granted I don't have a powerful machine dedicated to this so that may have been a big issue with the training session. Had I used a more powerful machine it may have taken significantly less time.

All in all I think the I biggest issue is having a large enough dataset for this task. Is it possible definitely since the question words were being classified as questions. This can be useful as you can then take the answers and send those to another nueral net that can maybe set a timer or set an alarm, or if it's a general question Google something. Maybe you need to know your next meeting. The other networks could check and return the final answer. 

Now we need to figure out how to best format these test cases into a gotcha section. 

I'm thinking initialization, dataset size, and data formatting. Since initialing the network a certain way would allow you to get better results, having a large enough dataset for something this complex is also needed ( and I still think I need to add more to the dataset I was using). And then data formatting was an issue e.g. having the input areay lengths the same and the output array lengths the same was important to yield actual numbers and not get NAN.

robertleeplummerjr:
A breathtaking ride!  Thanks for sharing!  Do you plan on sharing any of your code?  Please keep up the good work.  I'm working on recurrent neural nets in brian.js at the moment, and have been dabbling in taking advantage of multithreading so that training can be done a great deal faster, a minimum of 5 times and at the maximum around 700 times faster.

Kalevera:
Yea I'll work on posting my code this afternoon (hopefully) I'm going to be MIA through Thanksgiving with no internet or phones :) :D  I should be able to post it though since its already built out and only needs to be place on a repo. Maybe during lunch hmmmm.

robertleeplummerjr:
Quick point: The input and output indices do not have to be related.  Say if you had an input that varies of 1000 floating point numbers between 0 and 1 and the output you are trying to find something with a narrow output, simply decrease the outputs indices.  Example:

```js
var input = toIndices('What are all the letters in the world?');
var idealOutputs = {
  what: 0,
  who: 0.2,
  when: 0.5,
  where: 0.7,
  how: 1
};
```

You effectively could have one output, which would substantially improve training time.

robertleeplummerjr:
Too, I share your sentiment in regards to the node natural library.project(https://github.com/NaturalNode/natural/blob/5bbff4217b26cfaa1082cded2ae00e34b48c0bc6/lib/natural/util/stopwords.js#L25).  It is clever, but doesn't scale for multiple languages... _unless you built a neural net that could modify the natural library..._

Kalevera:
I initially did it that way. But then I wanted special words to arise like time,alarm, color, or timer this increased the output size. Also I was having trouble on how best to number them if im looking for a certain word? 

Or classifying the sentence as a declaration. 

The issue arises when you have array format with different output lengths as opposed to object format. I'm sure we could adapt the object I created to use object format as that seemed to be the cleaner and easier way of getting the data I wanted. I couldn't come up with a decent way of handling the data output. 

The array format I chose does the classification by converting the output back tona string. 

There's definitely room for improvement...

Kalevera:
Fyi I wasn't able to find time before leaving to push to a git repo. Will have to be Monday.

robertleeplummerjr:
Still excited to see results!

Kalevera:
Was able to upload my code I kept it separate from the brain.js files since its an object that I created and wasn't very fond of how I set it up.  I added a readme.md file its not very explanatory but I'm hoping the index-stream.js file is enough to understand what's happening. Will work on adding screenshots soon.

https://github.com/Kalevera/brainjs-classify-example

robertleeplummerjr:
A lot has happened since you opened this issue, the recurrent neural net is now working.  Now, I'm told (and do believe) that there are many better types of networks that would suffice for training similar to what you are wanting, but it may be the case that the recurrent neural net (specifically the LSTM or GRU) would be more ideal, or perhaps even the perfect fit here, other than the ANN (which I thought was just a standard Feedforward net until today, say hello to my little friend: https://github.com/harthur-org/brain.js/blob/1b92c81b0a5187712015d1eff7c941c50fce5a70/lib/neuralnetwork.js#L138) that is the default neural net in Brain.js, may be more ideal.

The reasons are that you want a somewhat flexible input and a specific output, a true recurrent neural net does just this, it doesn't get close, it gives you exactly that, sort of like a translation system.  Also, I feel that the input/output with a recurrent neural net is somewhat misunderstood...

Expounding:
The recurrent neural net is fed one character at a time, building up a memory of context, until the whole input is within the network, also you give it the expected result, or output...  Read that again and return here.  You read that correctly, the input and output are sent INTO the network in its training phase as the input.  The network is trained to know the difference by a specific index that is only used when separating an input from an output.  Then when predicting, or running the output, you simply feed the network the input followed by the separation character, and then run the prediction.  Now it may take some time to get training down, but most likely MUCH less time than with a feedforward neural net.

The reason it takes less time?  The neural net knows the entire state of the neural net, values can sort of intercommunicate with each other, and the network has context, whereas the ANN is more static mathematical calculations.

I'm working on the training method now, which _should_ accept your inputs just as they are in the library you posted, hang tight, I'll see if I can get it ready in a day or two.

I'm still sort of wrapping my head around the whole thing, and even some data scientists (http://karpathy.github.io/2015/05/21/rnn-effectiveness/) are as well, but hopefully this has helped enlighten the path a little.

How well is the library working for your results by the way?

Kalevera:
updated my readme.md to include the screenshots i took.

robertleeplummerjr:
Very nice work and dedication!  I did have some time to add a pseudo input/output to the rnn's, if you'd like to try them: https://github.com/harthur-org/brain.js/tree/rnn-train-pattern

You'd do something like this:
```
import LSTM from '../../src/recurrent/lstm';
import Vocab from '../../src/utilities/vocab';
var net = new LSTM({
  vocab: Vocab.allPrintableSeparated()
});
net.train([
  {
    input: 'hi',
    output: 'mom!'
  }, {
    input: 'hello',
    output: 'dad!'
  }
]);

//rnn prediction
console.log(net.run('hi'));
console.log(net.run('hello'));
```

Something I think too that needs addressed that could really help here and elsewhere is an auto adjustment of learning rate based off error rate.  Also, the error rate from rnn is different from that of the ANN, not sure honestly what the base value should be, but this should be enough for you to play with.

robertleeplummerjr:
Also, any suggestions for simplifying the api would be fantastically appreciated.

robertleeplummerjr:
Closing for the time being.  If this sparks up more conversation, and becomes something merge-worthy, I can reopen.

PR #56, opened 2016-12-28
toExpresseion() method added

braindy:
This method converts net to mathematical expression and wraps it to a
function. It must much accelerate calculations. (Sorry for non-ES6 syntx :)


braindy:
It is 50-100 times faster than normal function :$

robertleeplummerjr:
fantastic work!  I'll do the cleanup.

PR #68, opened 2017-06-06
Rewrite utility functions using ES6 and functional programming concepts

abhisheksoni27:
I have used this in a lot of places because it is known to be faster than iterating over an empty array.
```js
[...Array(size).keys()]
``` 

`Array.isArray(collection)` works all the time and beats the `constructor` check since it's **JS** and the `constructor` property can be mapped to something else, hence failing the test.

I know this doesn't improve the speed by ten folds, but I hope it helps in the long run.

And I hope it doesn't violate the design philosophy of the project.

### DIFF
--- a/src/utilities/mse.js
+++ b/src/utilities/mse.js
@@ -1,8 +1,8 @@
 export default function mse(errors) {
   // mean squared error
   let sum = 0;
-  for (let i = 0; i < errors.length; i++) {
-    sum += Math.pow(errors[i], 2);
-  }
+  sum = errors.reduce((currentSum, next) => {
+    return currentSum + Math.pow(next, 2);
+  }, sum);
   return sum / errors.length;
 }
--- a/src/utilities/ones.js
+++ b/src/utilities/ones.js
@@ -1,8 +1,7 @@
 export default function ones(size) {
   if (typeof Float64Array !== 'undefined') return new Float64Array(size).fill(1);
-  let array = new Array(size);
-  for (let i = 0; i < size; i++) {
-    array[i] = i;
-  }
+  let array = [];
+  array.length = size;
+  array.fill(0);
   return array;
 }
--- a/src/utilities/randos.js
+++ b/src/utilities/randos.js
@@ -1,9 +1,10 @@
 import randomWeight from './random-weight';
+import zeros from './zeros'
 
 export default function randos(size) {
-  let array = new Array(size);
-  for (let i = 0; i < size; i++) {
-    array[i] = randomWeight();
-  }
+  let array = zeros(size); // Just to fill the array so that map function works
+  array = array.map((element) => {
+    return randomWeight();
+  });
   return array;
 }
--- a/src/utilities/range.js
+++ b/src/utilities/range.js
@@ -5,9 +5,8 @@
  * @returns {Array}
  */
 export default function range(start, end) {
-  let result = [];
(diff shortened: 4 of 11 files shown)


abhisheksoni27:
I ran the test-gru command on my local machine and it passed twice, then failed, then passed again. What's wrong?

robertleeplummerjr:
...it became self aware, and only passes when it feels like it.

robertleeplummerjr:
jk, testing is a little touchy is all.  Probably just needs to be written better.

abhisheksoni27:
Sounds like a teenager to me :laughing:

abhisheksoni27:
Also, I think I would work on the tests after going through the source.

robertleeplummerjr:
My only concern is that iterating using the functional methods may be a bit slower than a plain loop.  Are these methods mostly preference, or do they perform better?

SISheogorath:
The usage of functions is dangerous. In JS calling functions is always slower than using loops as long as you do the same number of operations in it.

For `randos` I did a quick (not perfect but working) benchmark for NodeJS: https://gist.github.com/SISheogorath/4ac21f6d35b8ecf0185453f56fa7c3c6

The results are clear:

```
...
Endtime for a size of 16777216 elements with randos1: 756ms
Endtime for a size of 33554432 elements with randos1: 1483ms


...
Endtime for a size of 16777216 elements with randos2: 373ms
Endtime for a size of 33554432 elements with randos2: 744ms
```

Notice `randos1` is the new function. `randos2` the current one.

robertleeplummerjr:
@SISheogorath I wouldn't say they are bad, just that the practice here is to try and be as performant as possible, while having a very clear set of instructions.  I welcome the thinking here, and value the pr, but I'd be more inclined to use it if the functional aspects were more performant.  I will say that I've been added to the http://gpu.rocks team, and will be bringing gpu functionality into brainjs, and with that, there will be a bunch of function calls :) :) :).

robertleeplummerjr:
that being said, I think just re-looking at the functions, I bet we could speed them up by at least 50% more.

abhisheksoni27:
@robertleeplummerjr I created a `jsperf` test case to check `randos`, and I am afraid there is a hefty difference between `for` and `map`.([Test: JSPerf Map vs For Loop](https://jsperf.com/map-vs-for-loop-brain1))  Does this mean the functional programming concepts, passing callbacks sorcery and co fail when the size is considerably large. Maybe that's why JS is very cautiously mentioned with Machine Learning?

On a completely different note: I am gonna open a new PR to include the `Array.isArray()`. Should I keep the tests?

robertleeplummerjr:
Getting into the quick inner guts of how a function works, and specifically map:
* a closure is created, which costs
* there is array like variable that is created for the arguments sent into the closure, which costs
* map creates array
* probably a bunch of other magic

vs a for loop, which doesn't have half of that, functions are just more expensive.

I would LOVE to have those tests, and yet another contributor.  Tests and improvements to performance: bring em' on!

abhisheksoni27:
Wow. You made some important points there. It'd be great to put this argument in an article or something. :laughing: 

> I would LOVE to have those tests, and yet another contributor. Tests and improvements to performance: bring em' on!

This is one of the most generous comments I have ever seen regarding PRs. 

The tests are on their way. :rocket:

robertleeplummerjr:
(Love it!)

PR #70, opened 2017-06-10
Unique elements in array using Set

abhisheksoni27:
Just like a Set is defined in Mathematics, it has the same rules in JS. It cannot contain duplicate elements, because it wouldn't make sense.

The `let setFromArray = new Set([Iterable : Array])` creates a Set from an Array, thereby removing any duplicate elements.

If you pass that `setFromArray` to the spread operator, it creates an array of the keys of the set(keys here are the elements of the Set), which is what we want.

```js
[...new Set(collection)]
```

It takes care of empty elements (`''`) as well; since they are not allowed, they are automatically removed.

A JSPerf test to confirm improved speed: [Unique Array elements using Set](https://jsperf.com/unique-array-elements-using-set)

Cheers!

### DIFF
--- a/src/train-stream.js
+++ b/src/train-stream.js
@@ -162,16 +162,11 @@ export default class TrainStream extends Writable {
 
 /**
  *
- * http://stackoverflow.com/a/21445415/1324039
+ * https://gist.github.com/telekosmos/3b62a31a5c43f40849bb
  * @param arr
  * @returns {Array}
  */
 function uniques(arr) {
-  let a = [];
-  for (let i=0, l=arr.length; i<l; i++) {
-    if (a.indexOf(arr[i]) === -1 && arr[i] !== '') {
-      a.push(arr[i]);
-    }
-  }
-  return a;
+  // Sets cannot contain duplicate elements, which is what we want
+  return [...new Set(arr)];
 }


robertleeplummerjr:
this is much faster in ff, but far slower in chrome, can some investigation be done as to why?

wagenaartje:
@robertleeplummerjr 

Slower in chrome? Using Set is faster for me in chrome:

jsperf test

abhisheksoni27:
@robertleeplummerjr Which version of Chrome are you using?(Idk if that's relevant, though.)

It's faster for me on Chrome, as @wagenaartje pointed out.

brain js

robertleeplummerjr:
I'm the odd one out...
image

abhisheksoni27:
The key difference being: `Mac OS X` 
Have no idea why. :thinking:

robertleeplummerjr:
either case, we have enough use cases where it is better to go ahead and merge it in.

abhisheksoni27:
@robertleeplummerjr It'd be great to find out the reason for the aberration, though.

robertleeplummerjr:
Just updated to 59.0.3071, problem gone.

robertleeplummerjr:
image

abhisheksoni27:
Sweeeet. Thanks, Google! :tada:

PR #75, opened 2017-07-21
Refine and vectorize logo

janwirth:
See #20

PR #76, opened 2017-07-21
Fix logo

wagenaartje:
Uses rawgit. View the fixed `README.md` [here](https://github.com/wagenaartje/brain.js/blob/df18567405a3ccf11bb1f58fe9fab27dbb601093/README.md).


janwirth:
Nice fix. much better than mine as suggested in my latest post to #75. @robertleeplummerjr I have the gh-pages fix ready but I think Thomas' one is much cleaner.

Something seems to be wrong with travis-ci though...

robertleeplummerjr:
Nah, it is just we need better testing.

janwirth:
New logo looks nice :)

PR #93, opened 2017-09-14
Use the most negative possible value in pool

jwenzler:
I believe this provides a more firm extreme for the initial value of `largestValue`.


robertleeplummerjr:
Ty @jwenzler!  Great addition!

PR #95, opened 2017-09-23
opts.learningRate for train-stream

carlhopf:
when using train-stream allow set learningRate


carlhopf:
just found out the global options in new NeuralNetwork({ learningRate }) apply to streams aswell... closing

robertleeplummerjr:
If you find a better way, I'm all about it.

Issue #102, opened 2017-11-19
Faster training


arguiot:
Hello 👋, I would like to know if it would be possible to improve performances of the `train` function, using the GPU, or using multiple cores/threads in the node version.

Thank you for building this library

robertleeplummerjr:
The short answer: yes
The longer answer:
This is something we've put a great deal of thinking into, and ended up adopting [gpu.js](gpu.rocks) and assisting getting it to [version 1](https://github.com/gpujs/gpu.js/releases/tag/1.0.0).  I just [merged in an experimental branch](https://github.com/BrainJS/brain.js/blob/master/src/neural-network-gpu.js) this is where the above answer "yes" is satisfied.  That gives us our short term gain for feed forward neural networks.

The long term "yes" is that we are (re-?)structuring the recurrent architecture (which is still experiment) to create a single unified way of composing layers.  This branch: https://github.com/BrainJS/brain.js/tree/nn-gpu-layers embodies where we are going next.  Not only will it be fast (gpu), but easy, and straightforward.  This is something we feel is greatly missing from the machine learning community.  [The api is simple](https://github.com/BrainJS/brain.js/issues/99#issuecomment-352881840):

```js
const net = new FeedForward({
  inputLayer: () => { /* return a layer */ }
  hiddenLayers: [() => { /* return a layer */ },() => { /* return a layer */ },() => { /* return a layer */ }],
  outputLayer: () => { /* return a layer */ }
});
```

What is most important about this is that it will run on GPU first, with a CPU fallback, and that the Recurrent neural network will share the same basic architectural frontend, so no matter if you are building a simple perceptron, or a Vincent Van Gogh imagine manipulator, or a Johnny5 robot, or a cancer detector, you'll have the same easy means of composition.

I could go on and on and on, this has been far over a year in the making representing a lot of very smart people asking if the way current neural networks have been implemented right, and the answer we keep coming back to is "no", but I digress, and I thank you for asking this very fundamental question.

nsantini:
Ive been trying to get NeuralNetworkGPU to work but there seems to be a bug in it, see https://github.com/BrainJS/brain.js/issues/116

arguiot:
@nsantini Ok. I hope this bug will be fixed soon. I’ll try to investigate and if I find any solution to the problem I’ll answer in #116

robertleeplummerjr:
[...]

PR #121, opened 2018-01-13
Fix for #102 gpu exploding

robertleeplummerjr:
& bump version number


freddyC:
Just looking over it it seems fine, I haven't tested it yet.  I am not familiar with the `!!` used [here](https://github.com/BrainJS/brain.js/pull/121/files#diff-de1a5a7de991855714cdf433e9f8be96R1556) but it didn't change the functionality.

Any special notes on how I should be testing it?

robertleeplummerjr:
* For testing: That is a good point, the gpu stuff that is in v1 is currently beta, I had just altered the `likely` test, [here](https://github.com/BrainJS/brain.js/blob/e60c64eacd99c5ff9a29040e4b6c014ee0b30d73/test/base/likely.js#L37) to just use [the GPU class](https://github.com/BrainJS/brain.js/blob/5f48479a5ab0477c6e1b893e87df33a4b268c32f/src/neural-network-gpu.js) for the time being.  This resolves the issue, and kept me from going down the rabbit hole of splitting up and adding tests.  That'd likely need to be handled in another issue.
* for `!!` this switches logic twice, going from truthy, to false, and then to true.  I don't use it, and was just formatting that line to understand what was coming out of `toJSON` so to feed the gpu json into the cpu, and increment one time, and comparing.  It wasen't terribly fruitful, but in the end I liked the formatting change, so kept it.

freddyC (approved):
I think it looks good.  Running the leaky test got worked fine for me.  We do need more test coverage though (which you mentioned you wanted to do in a different issue).

Issue #98, opened 2017-09-27
net.run gives same response regardless of input


thegamecat:
I have a trained net that has an error of .06 (I've tested with an error of 0.0005) and when I do a netrun every single response is {"do":0.3964411059957613} regardless of what data is in input. 

Any ideas why?

robertleeplummerjr:
Can I see source?  I need more context.

thegamecat:
This is the stored nn:

{
  "layers": [
    {
      "ema20ema50percentage": {},
      "ema20ema100percentage": {},
      "ema20ema200percentage": {},
      "ema50ema100percentage": {},
      "ema50ema200percentage": {},
      "ema100ema200percentage": {}
    },
    {
      "0": {
        "bias": -1.5750287727890826,
        "weights": {
          "ema20ema50percentage": -20.25601029294507,
          "ema20ema100percentage": 0.45155990760160286,
          "ema20ema200percentage": 0.9709692038720795,
          "ema50ema100percentage": 20.639682793958684,
          "ema50ema200percentage": 19.975838224699498,
          "ema100ema200percentage": -1.4236833640338198
        }
      },
      "1": {
        "bias": 3.673306830993333,
        "weights": {
          "ema20ema50percentage": 30.770078299681398,
          "ema20ema100percentage": -5.425816420140965,
          "ema20ema200percentage": -4.227178255669408,
          "ema50ema100percentage": -36.613607664374,
          "ema50ema200percentage": -33.30627337398423,
          "ema100ema200percentage": 4.201335921604437
        }
      },
      "2": {
        "bias": 1.8227045570046092,
        "weights": {
          "ema20ema50percentage": 17.369191067278347,
          "ema20ema100percentage": -2.420666145569872,
          "ema20ema200percentage": -1.7769276688121898,
          "ema50ema100percentage": -20.369733465778662,
          "ema50ema200percentage": -18.42808116596809,
          "ema100ema200percentage": 2.3478189801860867
        }
      },
      "3": {
        "bias": 1.5185414425076638,
        "weights": {
          "ema20ema50percentage": 15.483940725383432,
          "ema20ema100percentage": -2.0763106966298435,
          "ema20ema200percentage": -1.5246698979511473,
          "ema50ema100percentage": -17.56431541365488,
          "ema50ema200percentage": -15.97979726318819,
          "ema100ema200percentage": 2.13326395858417
        }
      },
      "4": {
        "bias": 0.4936868611628317,
        "weights": {
[...]

PR #127, opened 2018-01-19
Fix multiple run issue

perkyguy:
If you run a NeuralNet (CPU or GPU) multiple times, with various inputs, then check the results after all the runs, the values will all be the same. e.g.
```js
const resultA = trainedNet.run([0,1,0]); // should be [0]
const resultB = trainedNet.run([1,0,1]); // should be [0.5]
const resultC = trainedNet.run([0,1,1]); // should be [1]

console.log(resultA, resultB, resultC) // [1], [1], [1]
```
This PR prevents this behavior by cloning the output layer's values in `runInput` rather than simply returning the output layer. The previous implementation (I believe) was breaking things because of object sharing.

Also, this standardizes the outputs of `run` for both `NeuralNetGPU` and `NeuralNet` to be an nontyped array. Previously GPU would return a texture, and CPU would return a typed array. This part was just out of simplicity from using the spread operator, rather than explicitly tying the result of `runInput` to a type.


perkyguy:
This may be the fix for https://github.com/BrainJS/brain.js/issues/98 , but without more context in that issue, it's hard to tell.

robertleeplummerjr (approved):
great work!

Issue #111, opened 2018-01-06
Quality Obj in toJson()


freddyC:
I think it would be beneficial to have a second item in the `toJson` object for recording quality of the network.  This will help when sharing networks with others.  to give them some kind of idea how many resources have been put into actually training that network.  It will also give some basic metrics in figuring out what types of networks take a really long time to train.

Perhaps something like this:
```
{
   layers: [ ... ],
   quality: {
      iterations: 70, // Times Trained
      accuracy: 0.0012323, // Percentage
      trainingSetLength: 15000,
      standardDeviation: 0.02 // no clue what actual numbers of this would look like yet
   }
}
```
  The idea for this is that I could create a network that is trained to determine handwriting, then I could give it to Lenny and he could continue training it from his data set that is different from mine, to improve it.  Then Carl could use it to solve his problem with Mo's terrible handwriting.

mubaidr:
https://github.com/BrainJS/brain.js/pull/150

PR #131, opened 2018-01-26
resolve lookup and set for importing values from json

robertleeplummerjr:
move toFunction into its own test, for easier debugging
add more tests, exposing issue
resolve issue which was expecting `layer` as an index, where `i` is actual index, and we need to set `this.biases[i][j]` which is a float, not overwrite it

### DIFF
--- a/test/base/json.js
+++ b/test/base/json.js
@@ -2,9 +2,9 @@ import assert from 'assert';
 import NeuralNetwork from './../../src/neural-network';
 
 describe('JSON', () => {
-  let net = new NeuralNetwork();
+  const originalNet = new NeuralNetwork();
 
-  net.train([
+  originalNet.train([
     {
       input: {'0': Math.random(), b: Math.random()},
       output: {c: Math.random(), '0': Math.random()}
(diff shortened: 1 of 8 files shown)


test/base/json.js
```diff
@@ -15,35 +15,74 @@ describe('JSON', () => {
...
-
-  it('toJSON()/fromJSON()', () => {
-    let output1 = net.run(input);
-    let output2 = net2.run(input);
-
-    assert.equal(JSON.stringify(output2), JSON.stringify(output1),
-      'loading json serialized network failed');
+  const input = {'0' : Math.random(), b: Math.random()};
+  describe('.toJSON()', () => {
+    describe('.layers', () => {
```
freddyC, line 23: Ticky-Tacky-Toey...I am not in love with the multiple describe lines, I don't mind them there but they kind of make the test output messy.  Look for comparison:

image


I don't hate it enough to ask you to change just enough to complain about it a little. (who really says `Ticky-Tacky-Toey` I mean come on)

robertleeplummerjr: lol
I don't prefer the output, but it is descriptive and allows us to compartmentalize the different items being exported both logically and for the human element.  I actually envisioned this expanding when more properties are potentially added here: https://github.com/BrainJS/brain.js/issues/111

freddyC (approved):
Looks good to me.  I complained about what bugged me but honestly it isn't a big deal, I am not going to request you change it I will just complain about it every time I get a chance until the good part of your heart feels guilty enough to change it........ or more likely I will probably forget about it.......until I have to do tests......

robertleeplummerjr:
fair enough

test/base/json.js
freddyC: energeticimpartialirrawaddydolphin-size_restricted
no-one-should-criticize-me

robertleeplummerjr:
fantastic gifs

Issue #135, opened 2018-01-28
Error given when forgotten to train


faroukcharkas:
## *What* is wrong?
When you forget to train the network an error message pops up saying something along the lines of `this.runInput is not a function`.

## *Where* does it happen?
This error appears when you run the neural network in the browser.

## *How* do we replicate the issue?
In order to replicate the issue, forget to train the network. Go from creation to running right away.

## *How* important is this (1-5)?
I would rate this a `1`. It'd just be great if it told you, "Hey, you forgot to train me!"

## Expected behavior (i.e. solution)
For it to instead of telling you, `this.runInp...` but instead say `Network was not trained on data` or something like that.

## Other Comments
I came across this when I forgot to train the network. Because of my forgetfulness, it costed me a day or so. This could've helped a little.

perkyguy:
This was fixed in PR #136 , if it's not, just re-open this and I can take a look at what happened.

robertleeplummerjr:
ty guys!

PR #136, opened 2018-01-29
Error logging for run() called prior to train()

perkyguy:
1621294-low_hanging_fruit_super

## Description

I added a function `isRunnable` as a `get` function, which will check for the side-effects of `train` being run. This 
In `run`, I added a call to 

## Motivation and Context

[issue](https://github.com/BrainJS/brain.js/issues/135)

## How Has This Been Tested?

Used the examples, and commented out tried doing a call to `run` prior to the `train` calls. Then `train`, and `run` again. The first time through, there should be an error message (and a `null` result), the second time through, it should work as expected. 

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


src/neural-network.js
```diff
@@ -119,12 +119,42 @@ export default class NeuralNetwork {
...
+    const checkFns = [
+      'sizes',
+      'outputLayer',
+      'biases',
+      'weights',
+      'outputs',
+      'deltas',
+      'changes',
+      'errors',
+    ].map(c => this[c]).every(fn => fn !== null);
```
robertleeplummerjr: If you change this to a filter, and check for length, you'd have the properties that failed

src/neural-network.js
```diff
@@ -119,12 +119,42 @@ export default class NeuralNetwork {
...
+
+    if(!checkFns){
+      console.error('Some settings have not been initialized correctly, did you run train()?');
```
robertleeplummerjr: which you could then use here, for a more specific error

src/recurrent/rnn.js
```diff
@@ -293,6 +293,21 @@ export default class RNN {
     this.ratioClipped = numClipped / numTot;
   }
 
+
+  /**
+   *
+   * @returns boolean
+   */
+  get isRunnable(){
+    if(this.model.equations.length === 0){
```
robertleeplummerjr, line 302: nice

src/neural-network.js
perkyguy: I like that much better. Check out the new implementation

src/recurrent/rnn.js
perkyguy: This is one of the defining side effects of `train`, correct?

src/neural-network.js
perkyguy: added, that is much cleaner/better

robertleeplummerjr (approved):
love it!

Issue #120, opened 2018-01-13
Private Functions in NeuralNetwork


freddyC:
So I have been thinking about https://github.com/BrainJS/brain.js/issues/99 and I think another thing that can get new users bogged down is how many methods and variables are exposed.  For example in the REPL I instantiate a simple NeuralNetwork and her are the things I have control over:

![image](https://user-images.githubusercontent.com/2294139/34903575-7999a272-f7f1-11e7-9841-e6eb5cc76467.png)

How many of these do we actually want the user to be interacting with?  If the user trains a network, then randomly calls initialize with some given size their network is `dust in the wind`.

I think if we take the methods that are meant to only be used internally within the class and declare them in the constructor (or some other scheme for having them not public).  I think that might also help those new to the library.  I don't mind doing this If you want me to.

  
### Just check boxes you want Private and I will take care of it.
- [ ] adjustWeights
- [ ] calculateDeltasLeakyRelu
- [ ] calculateDeltasRelu
- [ ] calculateDeltasSigmoid
- [ ] calculateDeltasTanh
- [ ] createTrainStream
- [ ] formatData
- [ ] fromJSON
- [ ] initialize
- [ ] run
- [ ] runInputLeakyRelu
- [ ] runInputRelu
- [ ] runInputSigmoid
- [ ] runInputTanh
- [ ] setActivation
- [ ] test
- [ ] toFunction
- [ ] toJSON
- [ ] train
- [ ] trainPattern
- [ ] activation
- [ ] biases
- [ ] binaryThresh
- [ ] calculateDeltas
- [ ] changes
- [ ] deltas
- [ ] errors
- [ ] hiddenLayers
- [ ] hiddenSizes
- [ ] learningRate
- [ ] momentum
- [ ] outputLayer
- [ ] outputs
- [ ] runInput
- [ ] sizes

robertleeplummerjr:
I'm hesitant to alter those items because of a few reasons:
* I would much rather we focus attention on v2 and v3, and utilizing ultra fast performance
* In javascript, there really isn't an a concept of private, it is more implied, most libraries just document the items they want supported in the wild, which we do, but could probably be enhanced
* the library has been established long before I ever became involved in it
* it can make future development more difficult
* the community often makes the best calls for this type of thing

So maybe if the community starts to respond: "Hey, yea, this has always bugged me, lets make it so!" then I'd be on board, but for now, I want to focus on innovation.

I'm really loving this energy!

[...]

PR #137, opened 2018-01-30
Options update

freddyC:

## Description

1) I updated the training options to be a property of the neural network.  You can now pass in any of the training options into the constructor and they will be recognized, set them to the network or pass them into the training and the property will get adjusted and trained appropriately.

2) I prepended underscores to neural network methods that are intended for private use

3) I did some general cleanup up with the new properties being on the class, and not passed into the methods.

4) I updated the tests to reflect changes. I also moved the logs test into their own file.

5) I included the training options in the saving of JSON so when the network is restored it's training options is restored as well.  If a log method is chosen it will restore to simply console.log.  The callback method is not restored.  This interaction is mentioned in the readme.

## Motivation and Context

[#111](https://github.com/BrainJS/brain.js/issues/111) will be helped but I don't think this finishes it
finishes [#120](https://github.com/BrainJS/brain.js/issues/120) for the neural-network class

The main focus is maintaining code.  This shouldn't change functionality other then extending it where logical.  It should help the library be simpler to implement as well as more robust and more maintainable in the future.

## How Has This Been Tested?

I added tests to mirror the functional changes.  I added in a test for passing things into the constructors.  I added tests for the training objects JSONification.

## Types of changes

- [ ] ~~Bug fix (non-breaking change which fixes an issue)~~
- [ ] ~~New feature (non-breaking change which adds functionality)~~
- [ ] ~~Breaking change (fix or feature that would cause existing functionality to not work as expected)~~
- [x] Maintenance and simplification

## Author's Checklist:

## Reviewer's Checklist:


test/README.md
```diff
@@ -12,7 +12,7 @@ Then you can run all tests using `npm test`.
 Run the unit tests with:
 
 ```
-npm test
+npm run test
```
robertleeplummerjr: I don't think you need `run` with newer npm.

robertleeplummerjr (changes requested):
can you remove `run`?

Fantastic work man!

src/neural-network.js
```diff
@@ -255,26 +263,66 @@ export default class NeuralNetwork {
...
+  /**
+   *
+   * @param options
+   *    Supports all `trainDefaults` properties
+   *    also supports:
+   *       learningRate: (number),
+   *       momentum: (number),
+   *       activation: ['sigmoid', 'relu', 'leaky-relu', 'tanh']
+   */
+  updateTrainingOptions(opts) {
```
perkyguy, line 295: This way of doing this seems a little cumbersome, especially if a new option wants to get added in. Exceptions to the standard (items contained in `this.trainOpts` are a little obfuscated as well. Maybe you could change this to something more like 
```
    Object.keys(NeuralNetwork.trainDefaults).forEach(opt => this.trainOpts[opt] = opts[opt] || this.trainOpts[opt]);
    this._setLogMethod(opts.log || this.trainOpts.log);
    this.activation = opts.activation || this.activation;
```

src/neural-network.js
```diff
@@ -255,26 +263,66 @@ export default class NeuralNetwork {
...
+    if (opts.timeout) { this.trainOpts.timeout = opts.timeout; }
+    if (opts.activation) { this.activation = opts.activation; }
+  }
+
+  /**
+   *
+   *  Gets JSON of trainOpts object
+   *    NOTE: Activation is stored directly on JSON object and not in the training options
+   */
+  _getTrainOptsJSON() {
```
perkyguy, line 313: Same comment as the one for `updateTrainingOptions`, we could change this to something like
```
return Object.keys(NeuralNetwork.trainDefaults)
      .reduce((opts, opt) => {
        if (this.trainOpts[opt]) opts[opt] = this.trainOpts[opt];
        return opts;
      }, {});```

**Review by robertleeplummerjr — APPROVED** [review]
:shipit:

File: src/neural-network.js [hunk 4]
```diff
@@ -255,26 +263,66 @@ export default class NeuralNetwork {
...
    */
-  _getSizesFromData(data) {
-    let sizes = [];
-    let inputSize = data[0].input.length;
-    let outputSize = data[0].output.length;
-    let hiddenSizes = this.hiddenSizes;
-    if (!hiddenSizes) {
-      sizes.push(Math.max(3, Math.floor(inputSize / 2)));
+  _verifyIsInitialized(data) {
+    if (this.sizes) return;
```
perkyguy, line 270: should this verify that the net was initialized correctly? Or just that at some point the sizes got set?

src/neural-network.js
```diff
@@ -255,26 +263,66 @@ export default class NeuralNetwork {
...
+   * If they are not it will initialize them based off the data set.
    */
-  _getSizesFromData(data) {
-    let sizes = [];
-    let inputSize = data[0].input.length;
-    let outputSize = data[0].output.length;
-    let hiddenSizes = this.hiddenSizes;
-    if (!hiddenSizes) {
-      sizes.push(Math.max(3, Math.floor(inputSize / 2)));
+  _verifyIsInitialized(data) {
```
perkyguy, line 269: This name should probably change, as it doesn't just verify that the initialization has occurred and will actually do something if needed. Not a big deal, as it's intended as an internal-only function, but still, might be better to refactor it a little to something more descriptive of it's actual function

src/neural-network.js
```diff
@@ -307,42 +360,39 @@ export default class NeuralNetwork {
...
+    const status = {
       error: 1,
       iterations: 0
     };
 
-    if (this.sizes === null) {
-      let sizes = this._getSizesFromData(data);
-      this.initialize(sizes);
-    }
+    this._verifyIsInitialized(data);
```
perkyguy, line 392: This is why I don't like this name...but I commented at the function definition.

src/neural-network.js
```diff
@@ -307,42 +360,39 @@ export default class NeuralNetwork {
...
 
-    while ( status.iterations < options.iterations && status.error > options.errorThresh) {
-      this._checkTrainingTick(data, status, options);
+    while (status.iterations < this.trainOpts.iterations && status.error > this.trainOpts.errorThresh && Date.now() < endTime) {
```
perkyguy, line 394: This PR is supposed to add in timeouts?

src/neural-network.js
```diff
@@ -351,41 +401,33 @@ export default class NeuralNetwork {
...
-      const items = new Array(options.iterations);
+      const items = new Array(this.trainOpts.iterations);
       const thaw  = new Thaw(items, {
         delay: true,
         each: () => {
-          this._checkTrainingTick(data, status, options);
-
-          if (status.error < options.errorThresh) {
+          this._trainingTick(data, status);
+          if (status.error < this.trainOpts.errorThresh || Date.now() > endTime) {
```
perkyguy, line 426: adding in timeouts?

perkyguy: Also, why is this after the tick now?

src/neural-network.js
```diff
@@ -307,42 +360,39 @@ export default class NeuralNetwork {
...
-  train(data, _options = {}) {
-    const options = Object.assign({}, this.constructor.trainDefaults, _options);
-    data = this.formatData(data);
-    options.log = this._setLogMethod(options.log);
-    options.learningRate = _options.learningRate || this.learningRate || options.learningRate;
-    var status = {
+  train(data, options = {}) {
+    this.updateTrainingOptions(options);
+    data = this._formatData(data);
+    const endTime = Date.now() + this.trainOpts.timeout;
```
perkyguy, line 385: adding timeout?

perkyguy:
This was kind of a large PR, with a little bit of scope creep

src/neural-network.js
freddyC: My idea was to just verify the size was set but it would probably be a good place to ensure the network was initialized and give the user feedback when they try to run without training.

src/neural-network.js
freddyC: This makes sense and is cleaner.  I am a little less familiar with the new JS magic.  The one thing I like about having the longer explicit if .... do this method is for functionality changes that may arise, then you don't have to re-parse to add functionality, but that is a weak argument.  Mostly the things that I am less familiar with are harder to read and I try and make it easy to read, doesn't mean it's the best lol.

src/neural-network.js
freddyC: The only place that gets weird is with the callbacks.  I have it set up right now where a callback doesn't get stored and log just stores true or false.  When restored the log method will be set as `console.log`  as long as we handle those this method seems much better.

src/neural-network.js
freddyC: yeah good point

src/neural-network.js
freddyC: yeah, so they can train for a set time interval as if want.

src/neural-network.js
freddyC: Yeah I completely forgot to add that in the pr, sorry.  I probably should have broken that out into it's own pr.

freddyC:
yeah sorry I'll try and keep things more concise

src/neural-network.js
perkyguy: the problem with the big `if (...) {} if (...) {}` was that if you want to add in a new option, you have to remember to add it into this list. All of the functions were doing the same thing, except for two exceptions (should those have even _been_ in `trainingOpts`?), so why not handle the same things the same?

src/neural-network.js
perkyguy: this is only for `log`?

freddyC: iirc the callback isn't stored but if log is a truthy value we store it as true, then when restoring it we restore it to the method `console.log`

PR #141, opened 2018-01-30
Making trainingOption updater/getter less rigid

perkyguy:
oil-man

## Description

I removed a lot of the hardcoded handling in `_updateTrainingOptions`, and instead have it iterate over all the `trainingDefault` options and set them if they exist. I didn't want to use a spread operator, as this keeps it limited to the options we want passed. this isn't a big deal IMO, and it could make this very simple to implement if we don't care about extra options.

## Motivation and Context

This is in response to some comments not being submitted before #137 was merged

## How Has This Been Tested?

`yarn test` - all passed

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


src/neural-network.js
```diff
@@ -292,17 +292,10 @@ export default class NeuralNetwork {
    *       momentum: (number),
    *       activation: ['sigmoid', 'relu', 'leaky-relu', 'tanh']
    */
-  updateTrainingOptions(opts) {
```
freddyC, line 295: I originally had this method as private (i.e. starting with `_`) but then asked myself why.  I think it is fine to be exposed to the user but I could see it going either way.  Mostly if we don't have the `_` it indicates to the user that they are safe calling it without having strange side effects.

I would vote for keeping it public but it isn't a big deal.

src/neural-network.js
```diff
@@ -292,17 +292,10 @@ export default class NeuralNetwork {
...
-    if (opts.log) { this._setLogMethod(opts.log); }
-    if (opts.logPeriod) { this.trainOpts.logPeriod = opts.logPeriod; }
-    if (opts.learningRate) { this.trainOpts.learningRate = opts.learningRate; }
-    if (opts.momentum) { this.trainOpts.momentum = opts.momentum; }
-    if (opts.callback) { this.trainOpts.callback = opts.callback; }
-    if (opts.callbackPeriod) { this.trainOpts.callbackPeriod = opts.callbackPeriod; }
-    if (opts.timeout) { this.trainOpts.timeout = opts.timeout; }
-    if (opts.activation) { this.activation = opts.activation; }
+  _updateTrainingOptions(opts) {
+    Object.keys(NeuralNetwork.trainDefaults).forEach(opt => this.trainOpts[opt] = opts[opt] || this.trainOpts[opt]);
```
freddyC, line 296: I think this would create weird behavior in the instance of updating training options.  For example:

```
net.train(data, { log: true });
net.train(data2, { logPeriod: 50 });
```
The second training would not print because the log would be reset to the default.  The same problem would occur any time the update training options method is called.

I like the idea of iterating instead of static if's but maybe we need to iterate on the union of keys from defaults and the passed in objects.

src/neural-network.js
```diff
@@ -311,22 +304,17 @@ export default class NeuralNetwork {
...
-    if (this.trainOpts.learningRate) { results.learningRate = this.trainOpts.learningRate; }
-    if (this.trainOpts.momentum) { results.momentum = this.trainOpts.momentum; }
-    if (this.trainOpts.callback) { results.callback = this.trainOpts.callback; }
-    if (this.trainOpts.callbackPeriod) { results.callbackPeriod = this.trainOpts.callbackPeriod; }
-    if (this.trainOpts.timeout) { results.timeout = this.trainOpts.timeout; }
-    if (this.trainOpts.log) { results.log = true; }
-    return results;
+    return Object.keys(NeuralNetwork.trainDefaults)
+      .reduce((opts, opt) => {
+        if (this.trainOpts[opt]) opts[opt] = this.trainOpts[opt];
```
freddyC, line 309: I don't think we want to add the json object of a callback to the results

freddyC (changes requested):
I like the changes.  There are a few things that I commented on that I think we need to handle still.  Let me know what you think.

robertleeplummerjr (dismissed):
options are now _fat free_!

src/neural-network.js
perkyguy: They set these options implicitly via the constructor or the `train`, do they need a third way to set the training options?

src/neural-network.js
perkyguy: I'm iterating over the keys of the static, not the static itself. I'm still setting the `this.trainOpts`

src/neural-network.js
robertleeplummerjr: I agree with this

src/neural-network.js
robertleeplummerjr: ideally this would throw an error, as the net is still training

src/neural-network.js
perkyguy: What _should_ it be? If someone wants to output it via `JSON.stringify`, the keys with functions in them will get omitted. This is a little awkward, but sending in some string representation of the function means we'd have to account for this on both converting to and from a JSON.  If they instead are just using the JSON as a holder in-memory for their net's configuration (e.g. how the tests use it), it would actually provide meaningful data back with references to the actual functions used rather than a placeholder.

src/neural-network.js
perkyguy: @robertleeplummerjr , I think this is assuming synchronous training, so there shouldn't be any overlap with train timing.

src/neural-network.js
robertleeplummerjr: Another question would be _why would we want to give the json of something someone else owns_?

robertleeplummerjr: Said differently:  There are likely values we do want to retain as json, perhaps we could just be static in defining them, rather than including all of them.  If it were, for example, just a list of the ones that matter, that'd give us both the small footprint of code, and logic for what actually affected the net, which may be important later for json exporting.

robertleeplummerjr: Following that train of thought, what are the values that actually affect the net?

robertleeplummerjr: Also, @perkyguy you know how I feel about `reduce`.


Only sith lords deal in `reduce`

src/neural-network.js
robertleeplummerjr: doh, that is correct.

robertleeplummerjr: all the new tools easily confuse me

src/neural-network.js
perkyguy: @robertleeplummerjr join me, and together we can rule the galaxy!

robertleeplummerjr: If we have a simple static list of the properties that affect the net that would later want to be referenced, it would be a static list and we'd simply use `map` on it.


p.s. @perkyguy I still see good in you, it is not too late

perkyguy: so the `callback` and `log` are the only two options which now could be "owned" by someone else (`log` since https://github.com/BrainJS/brain.js/pull/137/files#diff-f0964eec5cfcdf6b9c5b85748849be07R336). should we trim both of these out? Or just one? Or transform them to `'function() { ... }'`, or do their `toString()`?

src/neural-network.js
freddyC: no they don't need it, but if we have it already setup it is 0 cost to us letting it be explicitly exposed.  I am ok keeping it private I just thought I would bring it up.

src/neural-network.js
freddyC: SUPPER SORRY.  I read `... = opts[opt] || this.trainOpts[opt]` and thought `... = opts[opt] || this.trainDefaults[opt]`.  This is fine the way it is given that we call this in the constructor:

```
this._updateTrainingOptions(Object.assign({}, this.constructor.trainDefaults, options));
```

tenor

src/neural-network.js
freddyC: So I am of the opinion that we probably don't want to include an unknown function in the json (or even console.log).  I think it would be perfectly acceptable to either

1) convert log to true when set and restore it as console.log
2) Ignore both the `log` and `callback` properties of the training

PS I also like Array.reduce:
giphy

freddyC: One thing we don't address is validation to the trainOpts themselves.  We _should probably_ validate each of the values assigned.  Imagine learningRates in the negatives, or basically anything that is bad.

freddyC (approved):
So the first comment (the private vs public) I can see going either way, and I am fine with it.  I just thought I would mention it as it _might_ be convenient for users.  

The second (resetting training options) was my bad in reading it.  When I read it I thought "that will lead to weird behavior" but I totally read it wrong.  Sorry.

The third comment I think is the only thing left. I mentioned how I would vote to either ignore both or possibly restore `log` with the `console.log` method.  Either way we do it lets make sure the doc's reflect it.

The last thing we don't address is validation.  Do we want to?  If we do lets do that in a different pr (unless you want to group it I am fine either way).

robertleeplummerjr:
what kind of validation?

src/neural-network.js
robertleeplummerjr: nm, I see it now

robertleeplummerjr: My thought: using `reduce` is overkill.  Lets get a list of the items that should be specifically exported, then `map` them.

robertleeplummerjr (changes requested):
lookin' good, almost there

src/neural-network.js
robertleeplummerjr: it isn't that I hate `reduce`, it has it's place (dark side, unstructured data) and here we know the context and structure.

freddyC:
I think validation should be a new pr, I just brought it up (probably not the best place to do that but we are on the topic in lol).  Mostly with the focus of helping a user find out why their network isn't training.

src/neural-network.js
perkyguy: so are you proposing we have `get exportableTrainingOptions()` which will give us a subset of `get trainingDefaults`?

robertleeplummerjr: My initial thought was that this concept interfered with the general output would somehow make the library too non-specific enough for the lib, but the more I think about this, the more it makes sense.

ty guys for thinking this through, and it does clean up the training area.

robertleeplummerjr (approved):

robertleeplummerjr:
just fix up conflicts and we're good to go

PR #145, opened 2018-02-01
Make training methods DRY

perkyguy:
source

## Description

Added a new function `_prepTraining` which will do the basic things for initializing a training run, as well as provide a function for a training tick. Both `train` and `trainAsync` will use this tick function, either wrapped in thaw.js and a promise for `trainAsync` or simply run in a `while` loop for `train`. 

## Motivation and Context

No issue, just making this a little more manageable.

## How Has This Been Tested?

`yarn test` passed

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


src/neural-network.js
```diff
@@ -391,46 +392,54 @@ export default class NeuralNetwork {
...
+    const runTrainingTick = () => {
+      if (status.iterations < this.trainOpts.iterations && status.error > this.trainOpts.errorThresh && Date.now() < endTime) {
+        this._trainingTick(data, status);
+        return true;
+      }
+      return false;
+    };
+
+    return {
+      runTrainingTick,
```
freddyC: See this notation is new to me.  I am still getting the hang of the new js features, but this seems pretty cool.

src/neural-network.js
```diff
@@ -391,46 +392,54 @@ export default class NeuralNetwork {
 
     this._verifyIsInitialized(data);
 
-    while (status.iterations < this.trainOpts.iterations && status.error > this.trainOpts.errorThresh && Date.now() < endTime) {
-      this._trainingTick(data, status);
+    const runTrainingTick = () => {
```
freddyC: The only concern I can think of in making this a lambda is the overhead in the anonymous closure itself.  Generally I don't care about that but where this in the thick of the training, every iteration will be calling this so maybe we should care.  How would you feel about either putting this logic into the `_trainingTick` method or in it's own explicitly defined method?

src/neural-network.js
```diff
@@ -391,46 +392,54 @@ export default class NeuralNetwork {
...
     }
+  }
+  /**
+   *
+   * @param data
+   * @param options
+   * @returns {{error: number, iterations: number}}
+   */
+  train(data, options = {}) {
+    const { runTrainingTick, status } = this._prepTraining(data, options);
```
freddyC: again cool new notation.

freddyC (commented):
Looks good.  I have one comment on a lambda that I would like to hear @robertleeplummerjr comment on but I am not completely against how it is now.

freddyC:
I was thinking about it and saying `I am not completely against how it is now` probably gives off a more negative vibe then I was actually intending.  let me rephrase that to `I am fine with it merged the way it is, but I want to hear @robertleeplummerjr's opinion on the lambda`.

src/neural-network.js
robertleeplummerjr: In the sync training, lets not use a closure.  There really is no idea of a "tick", or a small allotment of "synthetic asynchronous" process.  As @freddyC mentioned, it is fairly heavy and can easily be avoided.

robertleeplummerjr: coming back to this as a pretty good idea, is there any way we can make it so we aren't relying on a closure for internal training?

robertleeplummerjr: thinking harder about this, I think it is a non-problem.  Consider:
_If_ we had 1,000 training sets, times 20,000 iterations, then we likely have something to care about for those that use this library, right.  Because 1,000 * 20,000 = 20,000,000 may add a couple seconds or so on modern js, but that isn't what is _dryed_ up here, this is the container for the training.

20,000 function calls isn't even a thing.  I say, this makes the js look good.  Lets roll it on in!

robertleeplummerjr (approved):
Looks good, ty ty!

robertleeplummerjr:
What do you say @freddyC?  It makes the code smaller, and makes things look a bit cleaner.  Also, there aren't any `array.reduce` usages...

src/neural-network.js
perkyguy: I fixed the closure issue you guys had. I just did something along the lines of what @freddyC said and moved the logic into `_trainingTick`, with that function returning whether or not it actually ran a train or not.

freddyC:
I was fine with the previous commit, I like this one even better.

freddyC (approved):

robertleeplummerjr:
:shipit:

Issue #146, opened 2018-02-02
Running brain.js inside a web worker


fede-ciliberti:
## *What* is wrong?
Hello, I am trying to run brain.js within a web worker in the same way as
the example in https://brain.js.org/ but for some reason the variable brain
is not being exported. Surely it is because the version has changed. I
appreciate if you can tell me how to proceed. Thank you very much.

robertleeplummerjr:
on it

robertleeplummerjr:
Resolved, and updated ph-pages so reflect it working: https://brain.js.org/

ty for finding this!

robertleeplummerjr:
You shiney new version: https://github.com/BrainJS/brain.js/tree/1.0.4

PR #147, opened 2018-02-02
Set globals for window, self, and module

perkyguy:
efecte_multipla

## Description

Added in a set to `self.brain` if it exists, will set either `window.brain`, `self.brain`, or `module.exports` if they exist.

## Motivation and Context

[issue](https://github.com/BrainJS/brain.js/issues/146)

## How Has This Been Tested?

load up a web-worker similar to the `gh-pages` branch, when you train a net it will throw an error. Update to this `browser.js` and starts working

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


robertleeplummerjr (approved):
fantastic!

Issue #142, opened 2018-01-31
User Object Validation


freddyC:
![A GIF or MEME to give some spice of the internet](https://media.giphy.com/media/SzkvLyuVgaGpW/giphy.gif)

## *What* is wrong?

We don't validate the training options that are passed in

## *Where* does it happen?

NeuralNetwork

## *How* do we replicate the issue?

Try passing in a negative timeout when training and see how well your network is trained.

## *How* important is this (1-5)?

2

## Expected behavior (i.e. solution)

We should at the very least log out when a value is not within a sane range saying `don't be stupid` or something polite like that.

## Other Comments
This probably isn't a big issue, but it's another low hanging fruit.

PR #148, opened 2018-02-03
Handles User Object Validation #142

freddyC:
Added in validation for training options and tests as well.

## Description

Creates validation for when the user sets the training options. 

## Motivation and Context

[issue](https://github.com/BrainJS/brain.js/issues/142)
This should help users who run into problems to see things they may have done wrong.

## How Has This Been Tested?

I created a slew of tests that go over the expected supported functionality

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


src/neural-network.js
```diff
@@ -35,12 +35,27 @@ export default class NeuralNetwork {
     };
   }
 
+  static get trainingValidations() {
```
perkyguy: Do we need a `get` for this? Why not just define these in the validation method itself?

src/neural-network.js
```diff
@@ -35,12 +35,27 @@ export default class NeuralNetwork {
...
+    return {
+      iterations: (val) => { return typeof val === 'number' && val > 0; },
+      errorThresh: (val) => { return typeof val === 'number' && val > 0 && val < 1; },
+      log: (val) => { return typeof val === 'function' || typeof val === 'boolean'; },
+      logPeriod: (val) => { return typeof val === 'number' && val > 0; },
```
perkyguy: logPeriod only matters if you also defined `log` as well. I think these functions don't _have_ to be a one-to-one mapping for the training options and can instead explicitly check the training options they are testing. This way, they all would be of the form `(opts) => { ... }`.

src/neural-network.js
```diff
@@ -35,12 +35,27 @@ export default class NeuralNetwork {
...
+    }
+  }
+
   constructor(options = {}) {
     Object.assign(this, this.constructor.defaults, options);
     this.hiddenSizes = options.hiddenLayers;
     this.trainOpts = {};
     this._updateTrainingOptions(Object.assign({}, this.constructor.trainDefaults, options));
 
+    this._isStrict = true;
```
perkyguy: If we are going to have multiple validation options, this should be contained within an object that contains the options and this as well. If we are not going to have multiple validation options, this should be named to something more descriptive, e.g. `this._strictTrainingOptionValidation` or something

src/neural-network.js
```diff
@@ -60,6 +75,16 @@ export default class NeuralNetwork {
     }
   }
 
+  /**
+   * 
+   * validateTrainingOptions is set true as default
+   *      When true: setting a training option outside expected range will throw an error
+   *      When false: setting a training option outside expected range will log out and continue
+   */
+  get validateTrainingOptions() { return this._isStrict; }
```
perkyguy: Return an object? the name suggests there are multiple

src/neural-network.js
```diff
@@ -60,6 +75,16 @@ export default class NeuralNetwork {
...
+  set validateTrainingOptions(s) { this._isStrict = s; }
```
perkyguy: pass in an object?

src/neural-network.js
```diff
@@ -366,6 +392,22 @@ export default class NeuralNetwork {
     return true;
   }
 
+
+  /**
+   *
+   * @param options
+   * @private
+   */
+  _validateTrainingOptions(options) {
```
perkyguy: This can be `static`, don't think it matters, but it can be.

src/neural-network.js
```diff
@@ -366,6 +392,22 @@ export default class NeuralNetwork {
...
+    Object.keys(options).forEach(key => {
+      if (NeuralNetwork.trainingValidations[key] && !NeuralNetwork.trainingValidations[key](options[key])) {
+        let message = `[${key}, ${options[key]}] is out of noraml range`;
+        if (this._isStrict) throw new Error (message);
+        else console.log(message);
```
perkyguy: `console.error`? `console.warn`?

src/neural-network.js
```diff
@@ -366,6 +392,22 @@ export default class NeuralNetwork {
...
+  /**
+   *
+   * @param options
+   * @private
+   */
+  _validateTrainingOptions(options) {
+    Object.keys(options).forEach(key => {
+      if (NeuralNetwork.trainingValidations[key] && !NeuralNetwork.trainingValidations[key](options[key])) {
+        let message = `[${key}, ${options[key]}] is out of noraml range`;
+        if (this._isStrict) throw new Error (message);
```
perkyguy: If these settings are wrong, this will break training, right? Why not just always enforce `isStrict`?

src/neural-network.js
```diff
@@ -366,6 +392,22 @@ export default class NeuralNetwork {
...
+
+  /**
+   *
+   * @param options
+   * @private
+   */
+  _validateTrainingOptions(options) {
+    Object.keys(options).forEach(key => {
+      if (NeuralNetwork.trainingValidations[key] && !NeuralNetwork.trainingValidations[key](options[key])) {
+        let message = `[${key}, ${options[key]}] is out of noraml range`;
```
perkyguy: typo in "normal"

src/neural-network.js
```diff
@@ -366,6 +392,22 @@ export default class NeuralNetwork {
...
   }
 
+
+  /**
+   *
+   * @param options
+   * @private
+   */
+  _validateTrainingOptions(options) {
+    Object.keys(options).forEach(key => {
```
perkyguy: Why not iterate over the `trainingValidations` functions? Only care about things we care about?

test/base/bitwise.js
```diff
@@ -75,30 +75,30 @@ describe('bitwise functions async training', () => {
     let not = [{input: [0], output: [1]},
                {input: [1], output: [0]}];
     testBitwiseAsync(not, 'not', done);
-  }).timeout(10000);
+  }).timeout(15000);
```
perkyguy: why the change?

test/base/bitwise.js
```diff
@@ -75,30 +75,30 @@ describe('bitwise functions async training', () => {
...
 
   it('XOR function', (done) => {
     let xor = [{input: [0, 0], output: [0]},
                {input: [0, 1], output: [1]},
                {input: [1, 0], output: [1]},
                {input: [1, 1], output: [0]}];
     testBitwiseAsync(xor, 'xor', done);
-  }).timeout(10000);
+  }).timeout(15000);
```
perkyguy: Why the change?

test/base/bitwise.js
```diff
@@ -75,30 +75,30 @@ describe('bitwise functions async training', () => {
...
 
   it('OR function', (done) => {
     let or = [{input: [0, 0], output: [0]},
               {input: [0, 1], output: [1]},
               {input: [1, 0], output: [1]},
               {input: [1, 1], output: [1]}];
     testBitwiseAsync(or, 'or', done);
-  }).timeout(10000);
+  }).timeout(15000);
```
perkyguy: Why the change?

test/base/bitwise.js
```diff
@@ -75,30 +75,30 @@ describe('bitwise functions async training', () => {
...
 
   it('AND function', (done) => {
     let and = [{input: [0, 0], output: [0]},
                {input: [0, 1], output: [0]},
                {input: [1, 0], output: [0]},
                {input: [1, 1], output: [1]}];
     testBitwiseAsync(and, 'and', done);
-  }).timeout(10000);
+  }).timeout(15000);
```
perkyguy: Why the change?

test/base/trainopts.js
```diff
@@ -173,3 +173,98 @@ describe('async train() options', () => {
...
+    assert.throws(() => { net._updateTrainingOptions({ errorThresh: false}) });
+    assert.doesNotThrow(() => { net._updateTrainingOptions({ errorThresh: 0.008}) });
+  });
+
+  it('log validation', () => {
+    let net = new brain.NeuralNetwork();
+    assert.throws(() => { net._updateTrainingOptions({ log: 'no strings' }) });
+    assert.throws(() => { net._updateTrainingOptions({ log: 4 }) });
+    assert.doesNotThrow(() => { net._updateTrainingOptions({ log: false }) });
+    assert.doesNotThrow(() => { net._updateTrainingOptions({ log: () => {} }) });
```
perkyguy, line 196: is this a "valid" log function?

test/base/trainopts.js
```diff
@@ -173,3 +173,98 @@ describe('async train() options', () => {
...
+    assert.doesNotThrow(() => { net._updateTrainingOptions({ momentum: 0.8 }) });
+  });
+
+  it('callback validation', () => {
+    let net = new brain.NeuralNetwork();
+    assert.throws(() => { net._updateTrainingOptions({ callback: 'no strings' }) });
+    assert.throws(() => { net._updateTrainingOptions({ callback: 4 }) });
+    assert.throws(() => { net._updateTrainingOptions({ callback: false }) });
+    assert.doesNotThrow(() => { net._updateTrainingOptions({ callback: null }) });
+    assert.doesNotThrow(() => { net._updateTrainingOptions({ callback: () => {} }) });
```
perkyguy, line 234: Is this a "valid" callback?

perkyguy (changes requested):

test/base/bitwise.js
freddyC: I had a test that timed out the first time.  I am actually thinking we should get rid of the async tests as much as possible (now that the training ticks are shared)  They really delay the test suit

test/base/trainopts.js
freddyC: yep, it would be user defined.  (logging allows the user to either set true and it will console.log or set their own method

test/base/trainopts.js
freddyC: yep (it just does nothing with the passed in object, I am not enforcing the params)

src/neural-network.js
freddyC: sorry I didn't see your comments but yeah I got rid of the getter/setter.  Originally when they set this to false I was printing out a message showing what it would do, but I thought that overkill.

test/base/bitwise.js
perkyguy: definitely should change up how we test the async. however, that should be done in a separate PR and I'd say just revert this back to limit unnecessary changes of this PR.

test/base/trainopts.js
perkyguy: I don't understand why the `log` value can be user defined, that's the point of `callback`. Buuuuuut, I was a bit too hasty with this and the other "Valid callback" comment. This is fine with me.

src/neural-network.js
freddyC: I am not sure I completely understand your point.  I could couple them so that we only check things that are dependent on each other and I could set them to check the value of the already set property.  I chose not to because I wanted to keep each property decoupled.  I think that will make it easier to extend (if you change the function of one property you don't have to worry about additional side effects) and I also didn't want to check every property each time I validate, just the properties that are changed.  Does that make sense?  I am open to changing but I don't yet understand the benefit.

src/neural-network.js
freddyC: totally agree the name didn't fit.  I ended up changing it before I saw this comment.

src/neural-network.js
freddyC: I don't see how that name suggests an object is returned.  I did change it from being a getter/setter. Originally I was printing out the a log saying what was happening when you set it to false but I thought that was a bit overkill especially once someone understands already how it works.

src/neural-network.js
freddyC: yeah it was a setter/getter but I got rid of that (since there aren't any intentional side effects)

src/neural-network.js
freddyC: Good idea, there really is no need to have it on the object instead of on the class.

src/neural-network.js
freddyC: yeah I only wanted to check the things that actually are changing.  I could see it going either way (the cost isn't huge).

src/neural-network.js
freddyC: I thought it would let the class stay more flexible to allow the validation to be shut off (in the odd case there is some weird option that they legitimately want to allow).  I kinda wrestled with this, it's that balance between providing structure and allowing innovation/creativity.  I landed on this due to the goal of the ticket, to help those who are new figure out things that might be going wrong.  Enforcing unexpected behavior I felt didn't fall within that (even though I can't think of a reason why they would want to do that).

src/neural-network.js
freddyC: Good idea, I'll change it to console.warn

test/base/bitwise.js
freddyC: 👍

test/base/trainopts.js
freddyC: No worries.  The main idea is so that if the user has their own logging system (i.e. it writes to file or whatever) then they can keep the log to do just that, without having to use up their callback for that.

perkyguy: Meh, I don't think it would "use up" their callback. But this is a moot point, we can continue discussion outside of this PR.

freddyC:
@perkyguy I think this is ready for another look-see

src/neural-network.js
```diff
@@ -35,12 +35,40 @@ export default class NeuralNetwork {
...
+    });
+  }
+
   constructor(options = {}) {
     Object.assign(this, this.constructor.defaults, options);
     this.hiddenSizes = options.hiddenLayers;
     this.trainOpts = {};
     this._updateTrainingOptions(Object.assign({}, this.constructor.trainDefaults, options));
 
+    this.invalidTrainOptsShouldThrow = true;
```
perkyguy: why put this on the instance and not just always require valid options? Brain won't train correctly without these set?

src/neural-network.js
```diff
@@ -35,12 +35,40 @@ export default class NeuralNetwork {
...
+      errorThresh: (val) => { return typeof val === 'number' && val > 0 && val < 1; },
+      log: (val) => { return typeof val === 'function' || typeof val === 'boolean'; },
+      logPeriod: (val) => { return typeof val === 'number' && val > 0; },
+      learningRate: (val) => { return typeof val === 'number' && val > 0 && val < 1; },
+      momentum: (val) => { return typeof val === 'number' && val > 0 && val < 1; },
+      callback: (val) => { return typeof val === 'function' || val === null },
+      callbackPeriod: (val) => { return typeof val === 'number' && val > 0; },
+      timeout: (val) => { return typeof val === 'number' && val > 0 }
+    };
+    Object.keys(options).forEach(key => {
```
perkyguy: Only reason I would go over all the validation keys rather than the passed in objects, was to ensure validity of all options?

perkyguy (commented):
Questions more than anything

robertleeplummerjr:
@perkyguy you waiting for anything else here?

perkyguy (approved):

perkyguy:
I think I'm good with all of this. Fix the merge conflicts though.

perkyguy:
I think I'm good with all of this. Fix the merge conflicts though, and make sure you build prior to merging.

robertleeplummerjr:
nice

PR #149, opened 2018-02-03
Add common .npmignore to exclude development files

daffl:
## Description

This pull request adds a `.npmignore` file to exclude files from being published that should be in Git but not in the installation.

## Motivation and Context

Currently build configurations, test files and IDE settings are included in the published version (when installing via `npm install brain.js`):

screen shot 2018-02-03 at 10 44 03 am

Usually this is not a big deal but when building in the browser, the `.babelrc` file is being read by Babel and complains about missing plugins even though the published version is already transpiled. This will also make the installation size a little smaller.

## Reviewer's Checklist:

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,10 @@
+.editorconfig
+.jshintrc
+.travis.yml
+.istanbul.yml
+.babelrc
+.idea/
+.vscode/
+test/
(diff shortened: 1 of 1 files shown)


robertleeplummerjr:
ty!

PR #150, opened 2018-02-03
Training Quality Object

perkyguy:
## Description

Add in a deepClone utility, as I needed to copy a structured object in the `fromJSON` method. I create a deepClone of the passed in json prior to everything else, this way we don't need to worry messing with the JSON messing with our net.

I also added in a training quality object to the net, which gets imported/exported in JSON form. This will give information on how the `train` method performed on it's last run, as well as for any training sets used. 

## Motivation and Context

[issue](https://github.com/BrainJS/brain.js/issues/111)

## How Has This Been Tested?

Added in a new test for `.trainingQuality`, can run `yarn` test to see it work.

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


src/neural-network.js
```diff
@@ -783,8 +805,9 @@ export default class NeuralNetwork {
    * @param json
    * @returns {NeuralNetwork}
    */
-  fromJSON (json) {
-    this.sizes = json.sizes
+  fromJSON (userJSON) {
+    const json = deepClone(userJSON);
```
robertleeplummerjr: any reason we can't do something like `JSON.parse(JSON.stringify(`?

robertleeplummerjr: bump

perkyguy: I mean, we could...but that just _seems_ so hacky to me. Plus that benchmark test I ran seems like we could use the `deepClone` in spots where we need more performance than `JSON.parse(JSON.stringify(...))`

robertleeplummerjr: You are right, it does seem hacky, I guess what I'd want to address is I'd like like not to have to support code we don't have to, is there a lib that gives us this, or an internal practice that node provides, rather than doing it ourselves?

perkyguy: https://lodash.com/docs/4.17.4#cloneDeep
https://www.npmjs.com/package/clone-deep
https://www.npmjs.com/package/deepcopy

robertleeplummerjr:
I say lets us a simple enough library, that is specific to this requirement.  If we end up needing to a lot of the functionality that lodash provides, we can migrate to it, but for the time being lets just use a simple lib and that way we don't own it.

src/neural-network.js
```diff
@@ -366,6 +368,25 @@ export default class NeuralNetwork {
...
+   * @private
+   * @returns {{error: number, iterations: number}}
+   */
+  _updateTrainingQuality(status, data){
+    this.trainingQuality.iterations += status.iterations;
+    this.trainingQuality.error = status.error;
+    this.trainingQuality.trainingSets.push({
+      ...status,
+      trainingSetLength: data.length,
+      local: true,
```
robertleeplummerjr, line 385: What does local mean?

perkyguy: Probably need to rename this. It differentiates if a training set was used locally or not. So if you import a net, then all training that had been done previously is deemed non local

robertleeplummerjr:
so all training sets, period, will be saved on the net?

robertleeplummerjr:
couldn't that be optional?  It seems like some scenarios where we have millions of training items will potentially cause an issue.

perkyguy:
Its each training set, or each time you run `train`

perkyguy:
So if you're running train a million times, then it would be somewhat of a problem

robertleeplummerjr:
"training items" meaning a single training set with a million or more items

robertleeplummerjr (approved):
fix up conflicts, and ensure unit tests pass.  This looks good to me.

freddyC (approved):
:shipit:

freddyC:
giphy

@perkyguy @robertleeplummerjr I know I should have spoken up a long time ago but after adding [this](https://github.com/BrainJS/brain.js/pull/184#pullrequestreview-109565108) comment I think I disagree with adding the new dependency.  On most projects I would agree with it but with brain.js I'd say lets do the work in house and not increase dependencies.

robertleeplummerjr:
Too far from source.  Perhaps better suited as an add-on or extension?

Issue #154, opened 2018-02-14
Faster Testing


freddyC:
![tenor](https://user-images.githubusercontent.com/2294139/36222183-bedd5d6a-117d-11e8-92e4-d86f9a9d0335.gif)

## *What* is wrong?

Tests take forever

## *Where* does it happen?

The tests where we use the async training

## *How* do we replicate the issue?

1. yarn test
2. wait
3. wait
4. wait
5. wait
6. wait
7. get up
8. go make a sandwitch
9. eat your sandwitch
10. grab a dewski
11. sit back down
12. wait
13. read the dictionary
14. wait
15. be happy the tests passed

## *How* important is this (1-5)?

probably a 42

## Expected behavior (i.e. solution)

1. yarn test
2. blink
3. realize you missed the test's finishing while you blinked

## Other Comments
We don't need to train so many things via async now that the async training and the regular training are using the same code paths.

robertleeplummerjr:
Is this specifically on the `NeuralNetwork` class?

freddyC:
The tests that are slow are all on the neural network class yes

PR #155, opened 2018-02-16
# If applied, this commit will make life better for everyone

freddyC:
### Why is this change needed?
To help with peace on earth

### How does it address the issue?
takes test from 90 seconds to 4 seconds

## Description

I remove tests that use the `trainAsync` method.
I add tests that verify that `train` and `trainAsync` in the `NeuralNetwork` class both use the same internal methods.

## Motivation and Context

[issue](https://github.com/BrainJS/brain.js/issues/154)

## How Has This Been Tested?

`yarn test` seems to do the job.

## Types of changes

- [ ] ~Bug fix (non-breaking change which fixes an issue)~
- [ ] ~Breaking change (fix or feature that would cause existing functionality to not work as expected)~

## Author's Checklist:

## Reviewer's Checklist:


test/base/trainopts.js
```diff
@@ -74,102 +74,78 @@ describe('train() options', () => {
   });
 });
 
+describe('train() and trainAsync() use the same internal methods', () => {
+  
+  it(' _prepTraining()', (done) => {
+    let trainingData = [ { input: [0, 0], output: [0] } ];
+    var net = new brain.NeuralNetwork();
+    let tick = 0;
+    net._prepTraining = (d, o) => {
```
robertleeplummerjr: Why is this calling itself?  It stack is going to get pretty deep, pretty fast.

freddyC: The tests pass, don't ask!  All we care about are the 
image marks.

Actually I think I am doing it wrong.

robertleeplummerjr: lol

After talking privately, I believe sinon will be your guid here.  I look forward to seeing this mature.

test/base/trainopts.js
```diff
@@ -74,102 +75,108 @@ describe('train() options', () => {
...
+  function getSpy (net, func) {
+    sinon.spy(net, func);
+  }
+
+  function killSpy (net, func) {
+    net[func].restore();
+  }
+
+  it('_prepTraining()', (done) => {
+    getSpy(net, '_prepTraining');
```
robertleeplummerjr: Can you move setup and teardown to `before(() => {});` or `after(() => {});`?

test/base/trainopts.js
```diff
@@ -74,102 +75,108 @@ describe('train() options', () => {
...
-    let p1 = net
-      .trainAsync(data, { errorThresh: 0.2 })
-      .then (res => {
-        assert.ok(res.error < 0.2, `[res.error, ${res.error}] should have been less then 0.2`);
+describe('train() and trainAsync() use the same private methods', () => {
+  let trainingData = [{ input: [0, 0], output: [0] }];
+  let opts = { iterations:1 };
+  let net = new brain.NeuralNetwork();
+
+  function getSpy (net, func) {
```
robertleeplummerjr: because of how simple it is, can you just call `sinon.spy` and `.restore()` directly?

freddyC: but what if I want to kill the spy?

robertleeplummerjr (approved):
good stuff here!

Issue #160, opened 2018-03-04
legacy support: handle if `trainOpts` doesn't exist


robertleeplummerjr:
From: https://github.com/karpathy/convnetjs/issues/54#issuecomment-370259874

freddyC:
I think this will be solved with the next npm release

PR #162, opened 2018-03-05
Fixes issue 160

freddyC:
- [x] Verified changes
 - [x] Tests, coverage, ran, checked

## Description

Because `Infinity` is not valid json it gets serialized as null.  We didn't catch this anywhere because we were
1) Using specific values in the json tests (not the default `Infinity`) 
2) Not actually serializing the json object, just using it as a javascript object.

I updated the Neural Network class to handle this when deserializing.

I also updated the tests to both check defaults and serialize the value.

## Motivation and Context

[issue](https://github.com/BrainJS/brain.js/issues/160)

## Types of changes

## Author's Checklist:

- [] ~I have updated the documentation as needed.~

## Reviewer's Checklist:


src/neural-network.js
```diff
@@ -830,6 +830,9 @@ export default class NeuralNetwork {
       }
     }
     if (json.hasOwnProperty('trainOpts')) {
+      if (!json.trainOpts.timeout) {
+        json.trainOpts.timeout = Infinity;
```
robertleeplummerjr: The problem with the approach is that it modifies the object being sent into brain. This is where I would have went as well, but it may be more advantagoes to do something like this, so as to ensure we aren't changing any values:

```js
if (json.hasOwnProperty('trainOpts')) {
  this._updateTrainingOptions({
    ...json.trainOpts,
    timeout: json.trainOpts.timeout === null ? Infinity : json.trainOpts.timeout
  });
}
```

This allows us to handle the exact situation where timeout is parsed to null, and treated as infinity, but has no side effects of altering the object the is sent to us.

robertleeplummerjr (changes requested):

robertleeplummerjr (approved):

robertleeplummerjr:
🚀 it

Issue #172, opened 2018-03-24
add "clean" script in package.json


giuscri:
This should be remove `node_modules`, ... to have the same repository you had right after clone.

renanbastos93:
This is good idea, so is think simple of implement.

PR #184, opened 2018-04-02
Added clean script

mubaidr:
https://github.com/BrainJS/brain.js/issues/172

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "dist": "babel src --out-dir dist --source-maps",
     "browser": "browserify ./index.js -p licensify -o browser.js -s brain",
     "browser-min": "browserify ./index.js -p licensify -s brain -g uglifyify -o browser.min.js",
-    "make": "rm -fr ./dist && npm run dist && git add ./dist && npm run browser && npm run browser-min"
+    "make": "rimraf dist && npm run dist && git add ./dist && npm run browser && npm run browser-min",
+    "clean": "rimraf node_modules"
   },
   "main": "./index.js",
   "devDependencies": {
@@ -37,7 +38,8 @@
     "mocha": "^3.0.2",
     "sinon": "^1.17.6",
     "testee": "^0.7.0",
-    "uglifyify": "^3.0.2"
+    "uglifyify": "^3.0.2",
+    "rimraf": "^2.6.2"
   },
   "keywords": [
     "ai",


freddyC (changes requested):
Hi @mubaidr 

Thanks so much for putting this together.

One of the appealing things about `brain.js` is how few dependencies it has.  I think it would be better to do the same work just in a script and not add in a new dependency.   I think the `packages.json` can just run the script with node (like this) i.e. `"clean": "node rimraf.js"`

I recently did this very thing in a different project and here is what I came up with:
```
'use strict';

var fs = require('fs');

var node_modules = './node_modules';

function deleteFolderRecursive (path) {
  if( fs.existsSync(path) ) {
    fs.readdirSync(path).forEach(function(file,index){
      var curPath = path + "/" + file;
      if(fs.lstatSync(curPath).isDirectory()) { // recurse
        deleteFolderRecursive(curPath);
      } else { // delete file
        fs.unlinkSync(curPath);
      }
    });
    fs.rmdirSync(path);
  }
};

deleteFolderRecursive(node_modules);
```

I was only using it on a mac so I have only tested it in iOS but I think it should be platform independent.  Do you mind adding a script to do this work and getting rid of the new dependency?

Thank you again good sir for stepping in and adding this, and sorry if it seems hyper critical.  One of the big focuses for brain is simplicity and one way to help that is to reduce the number of dependencies.

robertleeplummerjr:
One could reason that since we're removing one directory the simplest means would be `rm -rf node_modules` or `rmdir node_modules /s`.

Not to sound dogmatic, but is there a reason specifically that a clean script is needed?  It feels a bit like solving something that has already been solved.

robertleeplummerjr:
But I gotta say, I love the speed at which this was added. That enthusiasm is especially what any successful project needs!

freddyC:
Yeah it doesn't have to be a script, I was just thinking something that is cross platform, but I am open to either approach.  Mainly I think there is benefit in keeping dependencies low.

Macil:
It is only a devDependency: if someone runs `npm install brain.js`, then rimraf doesn't get downloaded and used. (Also, rimraf happens to already be a transitive dependency of some of the other current devDependencies of brain.js (ie. babel-cli -> chokidar -> fsevents -> node-pre-gyp -> rimraf), so adding it to package.json doesn't currently result in anything more being installed in dev. It definitely should be listed in package.json if it gets used directly by brain.js though.)

robertleeplummerjr:
I think @AgentME raises a good point, we already use it.  @freddyC, your thoughts?

freddyC:
@AgentME Great point, I totally didn't realize it was just a dev dependancy.  I also didn't realize it is already in the dependency tree.  With that being pointed out to me I am totally on board.  Sorry I didn't realize that before hand.

@robertleeplummerjr  I changed my review to be fine with this.

freddyC (approved):

freddyC:
@mubaidr Also thank you again for your pr, sorry I was wrong in my original review.  Great work man.

mubaidr:
@freddyC @robertleeplummerjr @AgentME Sorry, I have not been able to comment on this, probably would have solved this issue earlier. But thanks. :) Its just a very very minor contribution :p

Issue #192, opened 2018-04-14
New Feature: Time Step Prediction


robertleeplummerjr:
![A GIF or MEME to give some spice of the internet](https://media2.giphy.com/media/AgHBbekqDik0g/giphy.gif)

## *What* is the feature?
Prediction of trends.  Stock market, weather predictions, trends, etc.  This would be very similar to the existing recurrent net that will be deprecated, but will serve as a baseline for 2.0 GPU push.  This will be included in v1 somewhere, as it is a simplification of the existing recurrent network.

## *How* important is this (1-5)?
5

Usage:
```js
import brain from 'brain.js';

const net = brain.recurrent.RNNTimeStep(options);
// or
const net = brain.recurrent.LSTMTimeStep(options);
// or
const net = brain.recurrent.GRUTimeStep(options);

net.train([
  [1,2,3,4,5],
  [5,4,3,2,1],
]);

net.run([1,2,3,4]) -> 5
net.run([5,4,3,2]) -> 1
```

robertleeplummerjr:
It should be noted that this network type _does not_ work by represented data.  This type of neural network will work directly with the data you send in.

RobertLowe:
@robertleeplummerjr another curiosity, how might one use this approach in aggregation with other time step layers, for example using 3 stocks in the same time period?

I'd like to select a time range of various time step data, I'll have a score applied time step event (say news polarity for example 0.0<->1.0 ), and predict from sets like that?

robertleeplummerjr:
The more context the net has, the better and I'm not sure that the net would know how to distinguish one from the other.  I think it'd be best to train those nets separately.  The up-side to that would be each net would be smart enough to tracks its progress, but it would not do well at adjusting based off context of other financials.  Example: gold goes up, dollar goes down.

If some sort of non-generalized approach was needed I suppose a custom solution could be built.

The LSTM version (which would be my recommendation to use) uses simple math to achieve its equation: https://github.com/BrainJS/brain.js/blob/develop/src/recurrent/lstm.js#L48 so by modifying that, we may be able to achieve the context you are wishing for.

robertleeplummerjr:
I thought about this and it hit me this morning, this is already somewhat supported.  This is how:

The time step prediction currently only accepts a single value at a time, also it only outputs a single value at a time.  By default the network is setup as:

```js
const net = new brain.recurrent.LSTMTimeStep({
  inputSize: 1,
  hiddenSizes: [20, 20],
[...]

Issue #138, opened 2018-01-30
[Help needed] Guess next trend on chart


elmondogri:
Hello, I am not sure this is the best place to post my problem but I can't find another.

I am using Brain.js since a few days and am completly new to Machine Learning so I might be missing something obvious.

I have a huge (20k+) set of prices in json, something like:
`[{time:555555555000, price: 5.5456}, {time:555555556000, price: 5.5454}, {time:555555557000, price: 5.5457}, ...]`

Now what I want is, based on the last X prices, let brain.js guess if the price will go up or down for the next point.

I am doing something like this:

```
const brain = require('brain.js');
var net = new brain.NeuralNetwork();
var json = require('./tmp/test'); // no need to add the .json extension
    var data = [];
     var x = 100;

    for (var i = 0; i < json.length - 1; i++) {
	if (i < x) {
		continue;
	}

	var newData = {
		input: [],
		output: [1, 0, 0]
	};
	for (var i2 = (i - x); i2 < i; i2++) {
		if (json[i2+1].price > json[i2].price) {
			newData.input.push(1);
		} else if (json[i2+1].price == json[i2].price) {
			newData.input.push(0.5);
		} else {
			newData.input.push(0);
		}
	}
	
	if (json[i + 1].price > json[i].price) {
		newData.output = [0,0,1];
	}
	else if (json[i + 1].price == json[i].price) {
		newData.output = [0,1,0];
	}

	data.push(newData);
}
```

So as you can see, input contains the last 100 prices (I have tested with 10 and 50 and get the same error), already sorted by timestamp, and the output will be either [1,0,0] if 101th price is lower, [0,1,0] if it's equal, [0,0,1] if it's higher.

Then it trains using: 

```
net.train(data, {
				log: true,
				iterations: 20 // (have tried with 10, 100, 1000, 5000, error rate is always > 0.09
			});
```

Then I try to run, so I take 100 value somewhere inside json (still sorted by timestamp, so we are using the same as training data), and push them inside input array, but when I run 

```
var output = net.run(input);
console.log(output);	
```

output is always like the next price is higher, for exemple: `[ 2.233065781373218e-15, 0.2800779938697815, 0.9999711513519287 ]` and this even if next price is equal or lower.

What am I missing? I know it can't be 100% accurate, but something must be wrong as I always have the same result, aka new price is higher.

Thank you for your time.

freddyC:
[...]

Issue #109, opened 2017-12-30
LSTM examples?


jrobinson01:
Following on the other issue I created #108 , I'm trying to teach an LSTM network to write a simple children's book. I'm getting odd behavior but really don't know what I'm doing to begin with. I'd love to get this example working and added to the readme for others to follow but am hitting lots of little roadblocks. Here's my code:

```javascript
const brain = require('brain.js');

const words = new Map();
words.set('00001', 'Jane');
words.set('00010', 'Spot');
words.set('00100', 'Doug');
words.set('01000', 'saw');
words.set('10000', '.');

const trainingData = [
  {input: [0,0,0,0,1], output: [0,1,0,0,0]},// Jane -> saw
  {input: [0,0,0,1,0], output: [0,1,0,0,0]},// Spot -> saw
  {input: [0,0,1,0,0], output: [0,1,0,0,0]},// Doug -> saw
  {input: [0,1,0,0,0], output: [0,0,1,0,0]},// saw -> Doug
  {input: [0,1,0,0,0], output: [0,0,0,1,0]},// saw -> Spot
  {input: [0,1,0,0,0], output: [0,0,0,0,1]},// saw -> Jane
  {input: [0,0,0,1,0], output: [1,0,0,0,0]},// Spot -> .
  {input: [0,0,0,0,1], output: [1,0,0,0,0]},// Jane -> .
  {input: [0,0,1,0,0], output: [1,0,0,0,0]},// Doug -> .
  {input: [1,0,0,0,0], output: [0,0,0,0,1]},// . -> Jane
  {input: [1,0,0,0,0], output: [0,0,0,1,0]},// . -> Spot
  {input: [1,0,0,0,0], output: [0,0,1,0,0]},// . -> Doug
];

const lstm = new brain.recurrent.LSTM();
const result = lstm.train(trainingData);
const run1 = lstm.run([0,0,0,0,1]);// Jane
const run2 = lstm.run(run1);
const run3 = lstm.run(run2);
const run4 = lstm.run(run3);

console.log(words.get('00001'));// start with 'Jane'
console.log(words.get(run1));// saw
console.log(words.get(run2));// Jane, Doug or Spot
console.log(words.get(run3));// .
console.log(words.get(run4));// Jane, Doug or Spot

console.log('run 1:', run1, typeof run1);
console.log('run 2:', run2, typeof run2);
console.log('run 3:', run3, typeof run3);
console.log('run 4:', run4, typeof run4);
```

The results in the console:
[...]

PR #193, opened 2018-04-14
Add Time Step Recurrent Neural Network

robertleeplummerjr:
Add basic implementation of time step recurrent neural network so we can baseline for 2.0 and provide some support for community's requests.

This will answer #192 

tumblr_mw2o47v2ga1s8a6wyo1_400

## Motivation and Context

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


robertleeplummerjr:
This will also provide an answer for #138

robertleeplummerjr:
This will also provide an answer for #109

src/recurrent/matrix/equation.js
```diff
@@ -61,6 +62,21 @@ export default class Equation {
     return product;
   }
 
+  copy(m) {
+    let product = new Matrix(m.rows, m.columns);
+    this.states.push({
+      left: m,
+      product: product,
+      forwardFn: copy,
+      // backpropagationFn: () => {
```
freddyC: Do you want to leave in the commented out code?

src/recurrent/rnn-time-step.js
```diff
@@ -0,0 +1,169 @@
...
+  createInputMatrix() {
+    this.model.input = new RandomMatrix(this.inputSize, 1, 0.08);
+  }
+
+  createOutputMatrix() {
+    let model = this.model;
+    let outputSize = this.outputSize;
+    let lastHiddenSize = this.hiddenSizes[this.hiddenSizes.length - 1];
+
+    //whd
```
freddyC, line 20: What does `whd` and `bd` mean?  Is this just a reference to other code values?

src/recurrent/rnn-time-step.js
```diff
@@ -0,0 +1,169 @@
...
+    equation.add(equation.multiply(model.outputConnector, output), model.output);
+    model.equations.push(equation);
+  }
+
+  /**
+   *
+   * @param {Number[]} input
+   * @returns {number}
+   */
+  runInput(input) {
```
freddyC, line 56: Where this method is not meant to be used by the library user, perhaps an `_` before this method name would be beneficial? (it's semicolons but I think helps clarity)

src/recurrent/rnn-time-step.js
```diff
@@ -0,0 +1,169 @@
...
+    this.runs++;
+    let model = this.model;
+    let errorSum = 0;
+    let equation;
+    while (model.equations.length < input.length - 1) {
+      this.bindEquation();
+    }
+    const outputs = [];
+
+    if (this.inputSize === 1) {
```
freddyC, line 66: Could we change this from:
```
if (...) {
  for (<loop A> ) {
    for (<loop B>) {
//      code block A
    }
  }
} else {
  for (<loop A> ) {
    for (<loop B>) {
//      code block B
    }
  }
}
```

to:

```
for (<loop A> ) {
  for (<loop B>) {
    if (...) {
//    code block A
    } else {
//    code block B
    }
  }
}
```

Man that took way longer then it should have to abstract it lol.  Again it's semicolans but I think it helps clarity (and it looks like the only line that changes is: `const error = output.weights[i] - next;` vs `const error = output.weights[i] - next[i];` unless I am missing something.

src/recurrent/rnn-time-step.js
```diff
@@ -0,0 +1,169 @@
...
+          outputs.push(output.weights);
+        }
+      }
+    }
+    //this.model.equations.length - 1;
+    this.totalCost = errorSum;
+    return errorSum;
+  }
+
+  runBackpropagate() {
```
freddyC, line 108: Again an `_` I think helps clarity

src/recurrent/rnn-time-step.js
```diff
@@ -0,0 +1,169 @@
...
+   * @returns {Number[]|Number}
+   */
+  run(input = [], maxPredictionLength = 1, isSampleI = false, temperature = 1) {
+    if (!this.isRunnable) return null;
+    const model = this.model;
+    while (model.equations.length < maxPredictionLength) {
+      this.bindEquation();
+    }
+    let lastOutput;
+    if (this.inputSize === 1) {
```
freddyC, line 130: Again we could combine these loops into 1 that has the logic in it for `if (this.InputSize === 1) {` to just contain the `runInput([input[i]]);` or `runInput(input[i]);`

Again Semicolons.

src/recurrent/rnn-time-step.js
```diff
@@ -0,0 +1,169 @@
...
+      return lastOutput[0]
+    }
+    return lastOutput;
+  }
+
+  /**
+   *
+   * @returns {Function}
+   */
+  toFunction() {
```
freddyC, line 151: Thank you for including this!! I love that it's stubbed out and throws, so it is obviously clear if ever used what went wrong, but when we are ready to implement it we already have it stubbed out and just need to finish it.

src/recurrent/rnn-time-step.js
```diff
@@ -0,0 +1,169 @@
...
+    throw new Error('not implemented');
+  }
+}
+
+RNNTimeStep.defaults = {
```
freddyC, line 156: Did we get this represented in the documentation anywhere? (probably did but I haven't gotten that far yet).

test/recurrent/lstm.js
```diff
@@ -110,7 +110,7 @@ describe('lstm', () => {
       assert(lastOutput.length > 0);
       assert.equal(dataFormatter.toCharacters(net.toFunction()()).join(''), lastOutput);
     });
-    it.only('can include the DataFormatter', () => {
```
freddyC, line 113: lulz

test/recurrent/rnn-time-step.js
```diff
@@ -0,0 +1,134 @@
+import assert from 'assert';
+import RnnTimeStep from '../../src/recurrent/rnn-time-step';
+import LSTMTimeStep from "../../src/recurrent/lstm-time-step";
+
+/* NOTE: TimeStep here is deprecated though being committed as something new, it is the first feature we want using
+ recurrent.js because it is simple one of the simplest recurrent neural networks and serves as a baseline to completing
+ the GPU architecture.   This test is written so as to create the baseline we can measure against.
+ We get this working, we have a baseline, we finish recurrent.js, we change the world.
```
freddyC: You are too optimistic here, can you word it a bit more gloomily? 
tenor

freddyC (changes requested):
I had to keep it gloomy because I have been so bad at contributing lately.

tumblr_ow8zh8oiur1wzypxlo1_500

I have some things that I pointed out that I think you will be fine with most of them.

Let me know what you think.

freddyC:
Oh, but also great job on this! I don't think I actually said it but I did think it.  You have done some awesome stuff here so sorry to be the rain on you're parade.  

tumblr_l9zuy5air81qcr7fqo1_r1_500

I also think I might be gloomy because I started reading [this](https://www.audible.com/pd/History/Ordinary-Men-Audiobook/B01N7XXQCR) book recently and makes me sad.

src/recurrent/matrix/equation.js
robertleeplummerjr: I did not actually mean to add this method, removing.

src/recurrent/rnn-time-step.js
robertleeplummerjr: I did mean to keep this, it reflects the same/similar function from the `RNN` class.  This code will be deprecate in v2, so I think changing it would either require it changed in both locations or to just leave it alone.  This was originally in recurrent.js, which was used as a foundation to eventually get to v2.  I think the less work the better, but for the right reasons, so I think leave it.

src/recurrent/rnn-time-step.js
robertleeplummerjr: I'd like to leave it as it is for the time being, this code will be deprecated soon, and this is just used for baselining the new GPU stuff, changing it means we are going to maintain it and we've been very vocal that this the recurrent classes were beta.  Also the deprecated argument.

src/recurrent/rnn-time-step.js
robertleeplummerjr: I'm not going to waist time changing this because it is going to be deprecated.

src/recurrent/rnn-time-step.js
robertleeplummerjr: perhaps, but deprecated.

src/recurrent/rnn-time-step.js
robertleeplummerjr: Nice.

src/recurrent/rnn-time-step.js
robertleeplummerjr: Will document that now.

test/recurrent/rnn-time-step.js
robertleeplummerjr: Could your give me an example of what you are looking for?

src/recurrent/rnn-time-step.js
robertleeplummerjr: done

robertleeplummerjr:
Since the items were suggests and the actual net worked and closely resemble the existing recurrent neural network and will be removed after GPU implementations from the `nn-gpu-layers` branch is brought up to specification, went ahead and merged to keep things moving along.

freddyC:
giphy
gj man

PR #196, opened 2018-04-16
added example of recommendation engine (favorite t-shirts)

LegendOfGIT:
My favorite color is chrome

## Description
An expample of a recommendation engine, build with brain,js

## Motivation and Context
Show other devs how to use brain.js in the real world

## How Has This Been Tested?
Manually ;)

## Types of changes
- added a folder in examples with example code for recommendation engine

## Author's Checklist:

## Reviewer's Checklist:


freddyC (approved):
Typically I would suggest "Let put this in a sister repo to keep brain.js small/clutter free" but I think I am changing my attitude a bit.

Also this is a super simple example, well done, and starts sets up a way to add additional simple examples.  

giphy

The only thing I can think of to improve this (off the top of my head) would be to add a readme to this example's directory with a brief explanation on this example, then link that examples page on the main readme.

I would prefer the readme bit, but if you don't want to add it in I am not against this being merged as is.

robertleeplummerjr:
good work guys!

PR #203, opened 2018-04-28
Added some docs to simplify getting started

CreativeCactus:
## Require('meme')

## Description
Added the fields from the `defaults` objects used in the constructors of the two network types documented in the example section - to an object which is not documented currently but appears to allow networks to be configured quite easily. This makes it much easier for people to get up and running.

## Motivation and Context
For those who want to get up and running quickly and only read `the monospace fonts` - it should be immediately apparent both how to configure the networks and how they behave by default.

## How Has This Been Tested?
Spellcheck.

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,15 @@
     
 # Examples
 Here's an example showcasing how to approximate the XOR function using `brain.js`:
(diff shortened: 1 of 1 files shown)


robertleeplummerjr:
Nice work!

PR #216, opened 2018-06-05
Support NeuralNetwork.toFunction() when relu or leaky-relu is the activation function

jcjolley:
Without the IIFE, calling toFunction() on a neural network with relu, or leaky-relu set as the activation function causes toFunction() to throw a "SyntaxError: Unexpected token var"

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/src/neural-network.js
+++ b/src/neural-network.js
@@ -885,9 +885,9 @@ export default class NeuralNetwork {
         case 'sigmoid':
           return `1/(1+1/Math.exp(${result.join('')}))`;
         case 'relu':
-          return `(function () { var sum = ${result.join('')};(sum < 0 ? 0 : sum);})()`;
+          return `(function () { var sum = ${result.join('')}; return (sum < 0 ? 0 : sum);})()`;
         case 'leaky-relu':
-          return `(function () { var sum = ${result.join('')};(sum < 0 ? 0 : 0.01 * sum);})()`;
+          return `(function () { var sum = ${result.join('')}; return (sum < 0 ? 0 : 0.01 * sum);})()`;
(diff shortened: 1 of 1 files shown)


src/neural-network.js
```diff
@@ -885,9 +885,9 @@ export default class NeuralNetwork {
         case 'sigmoid':
           return `1/(1+1/Math.exp(${result.join('')}))`;
         case 'relu':
-          return `var sum = ${result.join('')};(sum < 0 ? 0 : sum);`;
+          return `(function () { var sum = ${result.join('')};(sum < 0 ? 0 : sum);})()`;
```
robertleeplummerjr: This looks a little funny.  The functions don't return anything.

jcjolley: Whoops.  Good call.

robertleeplummerjr:
Adding a lambda in that location is going to be called a ridiculously ridiculous (see what I did there) amount of times.  I think it would be prudent to follow either the existing non-lambda means or putting a lambda in a reusable location.

jcjolley (commented):
That look better?

jcjolley:
@robertleeplummerjr Yeah, that will be called a ton of times.  The problem is, we can't declare 'var sum' inside of an array, which is where those end up, so we either need to find a way to inline the sum, or make a function out of it.

robertleeplummerjr:
I'll look into this later today.

robertleeplummerjr:
Resolved, but want to fix a couple issues before releasing: https://github.com/BrainJS/brain.js/pull/257

PR #273, opened 2018-09-14
Refactor Neural Networks

dogagenc:
Some cleaning

## Description

Hey! This is really great project and fun to play with. Thanks for doing!

This PR applies DRY principle on `src/neural-network.js` and some other refactoring.

## Motivation and Context
I just realised there are some code blocks that doing exactly same thing except calculation in one line on `src/neural-network.js`. They are calculation of activation functions and deltas depending on activation type. I also think that it's not easy to add a new activation function to current code base. This makes activations more maintainable by keeping them on a separate object and use functions there. 

I also changed activation insertion on `toFunction` method to keep functions in one place and to be able to add new functions instantly.

P.S: There are some `let` keywords that never assigned a new value after declaration. It's a good idea to change them `const`s to keep memory in optimal state. And ideally, a linter would be nice to have. I'm thinking about to continue refactoring and configure Eslint when i have time.

Best,
 
## How Has This Been Tested?
New tests not being added since this is refactoring. All current tests have passed.

## Types of changes

- [x] Refactor

## Author's Checklist:

## Reviewer's Checklist:


ahmetcanaksu:
Aman allahım bir türk ??

src/neural-network.js
```diff
@@ -874,25 +758,12 @@ export default class NeuralNetwork {
...
-        case 'relu':
-          return `var sum = ${result.join('')};(sum < 0 ? 0 : sum);`;
-        case 'leaky-relu':
-          return `var sum = ${result.join('')};(sum < 0 ? 0 : 0.01 * sum);`;
-        case 'tanh':
-          return `Math.tanh(${result.join('')});`;
-        default:
-          throw new Error('unknown activation type ' + activation);
-      }
+      return `(${availableActivations[activation].activationFn.toString()})(${result.join('')})`;
```
robertleeplummerjr, line 766: This is a nice improvement to the code, but this does introduce function calls, which are overhead where before we were directly using math.  Could there be a means of applying the functionless means here as well?

dogagenc: @robertleeplummerjr Thank you. 🙏 

The one thing that i came my mind is to replace `sum`s with `result...` on activation function body and return it as a string. But this would also make code hard to maintain since it's hard to predict what would be added as new activation function in the future. Could be really complex activation that would cause changes on design of method - availableActivations - itself. Using same functions here would save us from unpredicted bugs and hardcoding complex activations IMO.

AFAIK, IIFE is good friend of memory since they immediately being released by mark-and-sweep garbage collectors on JavaScript engine. It might not be as bad as you think but i can't argue with it unless i have some benchmark data. :) 

So, i'll give it a try and compare performances of functions generated by current version and my version. I'll let you know about results.

Do you have any other suggestion here?

robertleeplummerjr:
Closing for time being as this adds too much recursion for performance and is in conflict now with existing files.  Very much so like the clarity here though.

Issue #185, opened 2018-04-02
Cross validation support


dan-ryan:
![Make sure to validate results](https://media.giphy.com/media/d7nd6bfKtMi7Co8DTS/giphy.gif)

Looking at the docs I couldn't see anything about cross-validation built into brain.js. A way to input a test size out of your input data and set test error would be nice to have.

May I suggest adding options similar to [Neataptic](https://wagenaartje.github.io/neataptic/docs/important/train/).

robertleeplummerjr:
There is no cross validator that exists currently in brain, but I actually am needing one as well.  I think the proposal is simple and to the point, it'd be to simply add something like:

```js
network.train(trainingSet, {
  crossValidate :
    {
      testSize: 0.4,
      testError: 0.02
    }
});
```

robertleeplummerjr:
I'm actually dead wrong!  https://github.com/BrainJS/brain.js/blob/11324b7dead480577a41420da8cbd32c3cd06684/src/cross-validate.js  It does exist.  @mubaidr what do you think about evaluating unit tests on it and ensuring it gets properly documented?

mubaidr:
Sure, will do.

robertleeplummerjr:
As it exists, we currently have this as the api:

`brain.crossValidate(brain.NeuralNetwork, trainingData, opts, trainingOpts);`

Because of the 1. amount of arguments, 2. that this isn't a documented feature, and 3. the existing api, what if we 1. shortened arguments to a single one, and 2. provided a means of passing in either a constructor or a instance of the neural network.  Proposal:

```js
brain.crossValidate({
  constructor: brain.NeuralNetwork, // use this
  instance: new brain.NerualNetwork() // or use this, but not both
  data,
  trainingOpts,
  opts,
  folds
});
```

robertleeplummerjr:
The reasoning behind changes is that I'd like to get at the json from the networks, and also I'd like to ensure whatever network I'm getting was the highest performing.

robertleeplummerjr:
I think we want to have the winning network as a result.  Perhaps something like `keyNetwork`?

robertleeplummerjr:
I ended up finding a much more dynamic means by using a class and instance rather than using raw json and passing the classifier around.  It is up for code review, and is as well documented therein.

robertleeplummerjr:
[...]

PR #280, opened 2018-09-27
Fix options is undefined!

mubaidr:
https://github.com/BrainJS/brain.js/issues/185#issuecomment-425034794


src/cross-validate.js
```diff
@@ -116,7 +116,7 @@ export default class CrossValidate {
       let dclone = data.slice(0);
       let testSet = dclone.splice(i * size, size);
       let trainSet = dclone;
-      let result = this.testPartition(options, trainOpts, trainSet, testSet);
+      let result = this.testPartition(this.options, trainOpts, trainSet, testSet);
```
robertleeplummerjr: Nice find!  We could (should) simplify this.  We don't need to pass options to `testPartition` because it is part of `this`, can you remove here and reference there?

robertleeplummerjr: I take it back, `this.options` just needs removed in this case. I'll go ahead and commandeer this as I need it for work this morning.

robertleeplummerjr (approved):

mubaidr (commented):
Looks good!

robertleeplummerjr:
Ty sir!

PR #302, opened 2018-11-12
Update DataFormatter.fromJSON to include specialIndexes

### LINKED ISSUE #299: fromJSON does not work with LSTM
I'm trying to use a LSTM network to just feed messages from a discord server, as a fun experiment to see what it can come up with. The problem is when I save the data using `toJSON` and then restoring it, it seems like the data is incompatible with what `fromJSON` expects.

```ts
const network = new brain.recurrent.LSTM();
const data = await getDataSomehow()

network.fromJSON(data)
```

`getDataSomehow` loads the json file, parses it and returns the object.
After doing this, this error is put in console whenever `run` is called: 
image

Furthermore, when I try training with new data, if it encounters a character that was not in the original training data, it gives me `Unrecognized character`.

What am I doing wrong?

felangel:
Bug Fix: https://github.com/BrainJS/brain.js/issues/299 (#299)

## Description
Include `specialIndexes` when creating `DataFormatter` from JSON

## Motivation and Context
[issue](https://github.com/BrainJS/brain.js/issues/299)

## How Has This Been Tested?

Tested locally against examples in the issue.

Note: I did not find any tests for `DataFormatter` so I am unsure if it is part of the scope of this ticket to add DataFormatter tests.

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/src/utilities/data-formatter.js
+++ b/src/utilities/data-formatter.js
@@ -164,6 +164,7 @@ export default class DataFormatter {
     dataFormatter.characterTable = json.characterTable;
     dataFormatter.values = json.values;
     dataFormatter.characters = json.characters;
+    dataFormatter.specialIndexes = json.specialIndexes;
     return dataFormatter;
   }
 


robertleeplummerjr (approved):
Nice work!

PR #309, opened 2018-11-24
svg network rendering

geokontop:
# SVG image for a feedforward network, created client-side or server-side.

## Description

- In src/utilities directory,  file svg.js with function makeSVG(size, options), that takes as arguments the network size and display options to return an svg XML string that describes the network topology. The file can be read from both the browser or the node.js
- In test/utilities directory file svg.js . In it tests for the makeSVG function.
- Inside examples/svg an example. Two scenarios. 
  1. The first one , concerns a client-side implementation (file index.html). All javascript code executed on the browser.
  1. The second one concerns a server-side node.js api that when requested, provides the svg string as a response to the client. A node server must be started first, listening on port 3000, with **node index.js** from the examples/svg directory. Then from the browser request **localhost:3000/serversideSVG.html**
## Motivation and Context

It gives a visual representation of the network. Can be used to provide a better user experience.

### Test results
svg
    check the value returned when sane inputs are provided
      √ should return a string
      √ should return a string starting with "<svg"
      √ should return a string ending to "</ svg>"
    "size" input
      √ should not throw an exception when size is not an array
      √ should return false when size length is smaller than 2
    "options" input
      √ should not throw an exception when any options missing

Used the existing testing enviroment 

New code does not affect other areas of the code

Screenshot from the example. Index.html after render request
screenshot1

An SVG response representation example
dest5
## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Author's Checklist:

- [X] My code focuses on the main motivation and avoids scope creep.
- [X] My code passes current tests and adds new tests where possible.
- [X] My code is [SOLID](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)) and [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).
- [X] I have updated the documentation as needed.

## Reviewer's Checklist:
- [X] I kept my comments to the author positive, specific, and productive.
- [X] I tested the code and didn't find any new problems.
- [X] I think the motivation is good for the project.
- [X] I think the code works to satisfies the motivation.


robertleeplummerjr:
I think this is amazing work!  Is there any way that, rather than using a whole webserver, you can use the newly added file directly, so that the webrequests can simply be a functional call?  Also, can you use sinon and simply assert for unit testing, rather than introduce chai & chai-xml?  Also, chai and chai-xml, if they are absolutely needed, would be devDependencies, rather than direct dependencies.

geokontop:
Well thank you! I made some changes following your recomendations. Now the svg string can be constructed on the client side as well. The svg.js file is perceivable from both node.js and the browser.

**In the examples/svg**
**index.html** demonstrates the client aproach, where svg.js is downloaded and the makeSVG function is executed by the browser which subsequently renders the result. No need to start server.
**serversideSVG.html** demonstrates the serverside approach. In that case, the browser sends an AJAX request to the server and a node.js api takes care of constructing and returning the string. Need to start server with **node index.js** and request **localhost:3000/serversideSVG.html**

**In package.json**
I removed the chai and chai-xml. I had to skip a test for that. The output is not checked whether it is well formed xml.

robertleeplummerjr:
I think (and I could be mistaken) because the focus of brain.js is for a solid app in js and node that the _server side implementation_ (note: just that part, not the svg generation as a whole), although useful, is outside of the spectrum of what we want to exemplify.  We could open another project if you like, and possibly put it there.

As for the class that generates svg and validation on it being xml, do you have any suggestions other than just commenting out code?  We currently use sinon in master for any testing outside of `assert`.

Also, are there plans to support any of the other neural networks that are in brain, other than just `NeuralNetwork`?

geokontop:
@robertleeplummerjr , answering in reverse order

I would like to support the project in generall. SVG rendering for other types of ANNs (if that is what you mean) could be the next step. RNN specificaly, Also some improvements in current feedforward svg facilities.

I can't propose any other way than using chai-xml, if xml well-formed testing is considered crucial. In that case chai and chai-xml could be used as dev-depemdencies.

As for the serverside issue, a lot of people (and I am one of them) are inerested in using javascript in backend implementations. Node.js made that possible and that is what the example demonstrates. So as a first thought, why not show/give that potentiality as well. If that is considered short of "scope overloading" then I could remove the server-side branch in my example.

A node.js api project would be interesting anyway.

robertleeplummerjr:
>I would like to support the project in generall. SVG rendering for other types of ANNs (if that is what you mean) could be the next step. RNN specificaly, Also some improvements in current feedforward svg facilities.

I look forward to it!  Especially with the new GPU architectures coming.

>I can't propose any other way than using chai-xml, if xml well-formed testing is considered crucial. In that case chai and chai-xml could be used as dev-depemdencies.

There are numerous options available, one that is independent of the testing framework is https://www.npmjs.com/package/fast-xml-parser, for example.  The reason that `chai` and `chai-xml` aren't ideal is because we are currently using mocha to test.  Introducing another testing framework would only be confusing and unneeded at least in this version of brain.  I do agree, if we are going to introduce svg, it does need to be validated though.

>As for the serverside issue, a lot of people (and I am one of them) are inerested in using javascript in backend implementations. Node.js made that possible and that is what the example demonstrates. So as a first thought, why not show/give that potentiality as well. If that is considered short of "scope overloading" then I could remove the server-side branch in my example.

First off, I love your passion.  But I do want to interject.  You speak of brain.js as not being node.js, or it in a sense not being backend.  It is.  Node is.  It is also front end.  It is both.  It is built node.js first, so that it can be packaged, repackaged, unpackaged, browserified, or uglyfied for use wherever you like: front, or back end.  Because that is generally understood and accepted, it is as well understood that because you can call `instance.method()` here (client?) you can as well call `instance.method()` there (server?). 

The `examples` (and as well the `examples-typescript`) folder are to expound on examples of what are specific to the usages of what are defined within this library.  And because the server side functionality exist outside of this library (spool up a server, so that I can make a request for an html page, that can make a request, to eventually call `instance.method()` so that I can display that on the client) we can just demonstrate the `instance.method()` calls directly.  The developer implementing the library can decide how or where that may go.

>A node.js api project would be interesting anyway.

I agree, and that has been my goal with this project.  My focus, aside from (wife, kids, job,) code reviews, and bug fixes, and features that the community wants, has me pushing to make node.js's GPU capabilities second to none.  That is why I've been working on gpu.js, but that is a horse of a different color.

Code review:

* Here (https://github.com/BrainJS/brain.js/pull/309/files#diff-3ffc1f0282ccc977e754634520cc4c85R8) you've used the convention "makeSVG", would you mind changing it to match our existing "to" methodology that we're following from node.js?  Example rather than "makeSVG" simply "toSVG".
* This method (https://github.com/BrainJS/brain.js/pull/309/files#diff-3ffc1f0282ccc977e754634520cc4c85R8) uses the arguments `size, options`, why not simplify it so the net can be passed in?  Like this: `toSVG(network, options)` or even `toSVG(json, options)`, that way the implementer doesn't have to think about what options need to be sent in, but rather just uses the network as a means of making the svg.
* Here (https://github.com/BrainJS/brain.js/pull/309/files#diff-3ffc1f0282ccc977e754634520cc4c85R8), and the lines following, the logic you are using for deducing what options have been selected and the defaults is quite verbose, it could greatly be simplified by using a default object, much like this: https://github.com/BrainJS/brain.js/blob/bee2042a49519094ac9ef01f56777436c3072abb/src/neural-network.js#L72
* Here (https://github.com/BrainJS/brain.js/pull/309/files#diff-3ffc1f0282ccc977e754634520cc4c85R68) you've made it so that the utilities can be both a global (window) export or a node.js export.  While that is valuable, there is already a namespace associated with brain.js, that can easily be used externally referencing.  Example - (https://github.com/BrainJS/brain.js/blob/bee2042a49519094ac9ef01f56777436c3072abb/index.js#L41 which woudl give us something like: `brain.utilities.toSVG()`
** There is, however, a `likely` utility, that if the above were to have a network input, would fit well taking a similar approach.
* Missing semicolons

geokontop:
- [X] function **toSVG** filename **to-svg** in src/utilities
- [X] argument **network** instead of **size**
- [X] function called as brain.utilities.toSVG{network,options) 
- [X] verbosity reduced
- [X] dist files, browser.js, browser.min.js
- [X] fast-xml-parser as devDependency, testing valid xml
- [X] semicolons
- [x] example implementation both clientside and serverside API call
- [ ] documentation
- [ ] still verbose, default object for network?
- [ ] is testing adequate?

**Edit:** Great ... conflicts!

geokontop:
I couldn't handle the conflicts. I will request a new pull.

PR #324, opened 2018-12-26
Fix broken link in readme

ajthinking:
## Description
Could not click through to the contributors graph because github adds "master/blob" even when using relative paths. Applied full url instead.

Thanks


robertleeplummerjr:
Nice gif!

PR #349, opened 2019-01-23
fix: Ensure property is own when iterating over nodes object

shama:
This ensures when iterating over `nodes` within `fromJSON()` with `for..in` that it will only iterate over properties the object owns and not other unintended properties on the prototype.

## Motivation and Context
By default, Ember.js extends all kinds of native prototypes. So when using `fromJSON()` within a default Ember app, it tries to iterate over the prototype properties Ember adds. Such as `_super`.

Without this change, Ember apps get the error `Uncaught TypeError: Cannot read property 'bias' of undefined` because eventually it hits the Ember prototype property: `this.biases[i][j] = layer['_super'].bias;` which doesn't have the `bias` property.

## How Has This Been Tested?
I modified the `node_modules/` in my app.

screen shot 2019-01-22 at 5 40 25 pm

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/src/neural-network.js
+++ b/src/neural-network.js
@@ -982,9 +982,11 @@ export default class NeuralNetwork {
         const nodes = Object.keys(layer);
         this.sizes[i] = nodes.length;
         for (let j in nodes) {
-          const node = nodes[j];
-          this.biases[i][j] = layer[node].bias;
-          this.weights[i][j] = toArray(layer[node].weights);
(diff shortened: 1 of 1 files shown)


robertleeplummerjr (approved):

robertleeplummerjr:
Nice work.

PR #355, opened 2019-02-04
Fix #354: Update learn-math example to fix learning data and calculate error on test

### LINKED ISSUE #354: Example learn-math.js/learn-math.ts training data looks weird
## *What* is wrong?
Training data looks wrong, not a big deal to train a model with wrong data if we expect wrong results, but that's a bit awkward when you look at the example, see [there](https://github.com/maxme/brain.js/blob/master/examples-typescript/learn-math.ts#L116):
```
 '8+9=18',
 '9+8=18',
 '9+9=18'
```

## *How* do we replicate the issue?
Look, run the `learn-math.[js|ts]` example.

## *How* important is this (1-5)?
1

## Expected behavior (i.e. solution)
In my book ;) :
```
 '8+9=17',
 '9+8=17',
 '9+9=18'
```

## Other Comments

Would be great to update the test with unknown data, eventually print error instead of a list of booleans.

maxme:
Fix #354: Update learn-math example to fix learning data and calculate error on test

### DIFF
--- a/examples-typescript/learn-math.ts
+++ b/examples-typescript/learn-math.ts
@@ -18,9 +18,9 @@ const net = new LSTM();
 const mathProblems = [
   '0+0=0',
   '0+1=1',
-  '1+0=2',
+  '1+0=1',
   '0+2=2',
-  '2+0=3',
+  '2+0=2',
   '0+3=3',
   '3+0=3',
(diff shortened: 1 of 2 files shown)


robertleeplummerjr:
Ty very much for finding this!

PR #361, opened 2019-02-16
fixed "this.log" => "console.log"

xonoxitron:
Errors: 
+ this.log('This iteration\'s data length was different from the first.');
+ this.log(`iterations: ${ this.i}, training error: ${ error }`);

Why?
"this.log" is not a function, it's boolean (inherited from "trainOpts.log")

### DIFF
--- a/src/train-stream.js
+++ b/src/train-stream.js
@@ -83,7 +83,7 @@ export default class TrainStream extends Writable {
    */
   finishStreamIteration() {
     if (this.dataFormatDetermined && this.size !== this.count) {
-      this.log('This iteration\'s data length was different from the first.');
+      console.log('This iteration\'s data length was different from the first.');
     }
 
     if (!this.dataFormatDetermined) {
@@ -100,7 +100,7 @@ export default class TrainStream extends Writable {
     const error = this.sum / this.size;
 
     if (this.log && (this.i % this.logPeriod === 0)) {
-      this.log(`iterations: ${ this.i}, training error: ${ error }`);
+      console.log(`iterations: ${ this.i}, training error: ${ error }`);
     }
     if (this.callback && (this.i % this.callbackPeriod === 0)) {
       this.callback({


robertleeplummerjr:
This is incorrect.  This.log, is being checked for truthy.  It is a boolean _or_ a function.

robertleeplummerjr:
But also, ty for your hard work investigating.  I grappled with this when I first looked at the code as well.

xonoxitron:
Hi Robert,
according to this:

*Training Options*
train() takes a hash of options as its second argument:

net.train(data, {
                            // Defaults values --> expected validation
      iterations: 20000,    // the maximum times to iterate the training
data --> number greater than 0
      errorThresh: 0.005,   // the acceptable error percentage from
training data --> number between 0 and 1
*      log: false,           // true to use console.log, when a function is
supplied it is used --> Either true or a function*
      logPeriod: 10,        // iterations between logging out --> number
greater than 0
      learningRate: 0.3,    // scales with delta to effect training rate
--> number between 0 and 1
      momentum: 0.1,        // scales with next layer's change value -->
number between 0 and 1
      callback: null,       // a periodic call back that can be triggered
while training --> null or function
      callbackPeriod: 10,   // the number of iterations through the
training data between callback calls --> number greater than 0
      timeout: Infinity     // the max number of milliseconds to train for
--> number greater than 0
});

it is misaligned with this:

// inherit trainOpts settings from neuralNetwork
neuralNetwork.updateTrainingOptions(options);
const { trainOpts } = neuralNetwork;
this.iterations = trainOpts.iterations;
this.errorThresh = trainOpts.errorThresh;
*this.log = trainOpts.log;*
this.logPeriod = trainOpts.logPeriod;
this.callbackPeriod = trainOpts.callbackPeriod;
this.callback = trainOpts.callback;

so, if *this.log *is *false OR a method *you are right, but if it is *true*,
you are *wrong* because it should use *console.log.*

In my pull request the case was:

this.log = true


so the following lines:

this.log(`iterations: ${ this.i}, training error: ${ error }`);
and
this.log('This iteration\'s data length was different from the first.');


give error.

happy coding buddy!

M

robertleeplummerjr:
Could you abstract this function for reuse perhaps, and reuse?: https://github.com/BrainJS/brain.js/blob/8455336e274277f571b236d98732b9f0890f5e63/src/neural-network.js#L368

xonoxitron:
The method is abstract enough, but in train-stream you have the neural
network object as property.
Inside the neural network the method is called, outside not.

const { trainOpts } = neuralNetwork;

should be

const { trainOpts } = neuralNetwork.trainOpts;

in a case like:

const config = { binaryThresh: 0.5, hiddenLayers: [3], activation:
'sigmoid', leakyReluAlpha: 0.01, trainOpts: { log: true } }; const net =
new brain.NeuralNetwork(config);

:)

PR #373, opened 2019-03-13
Adding example for LSTM with multiple output

acagastya:
Added an example for LSTM for determining if the test string starts with "A", "B", or "C".


mubaidr (approved):

mubaidr:
@acagastya thanks

PR #401, opened 2019-05-31
resolves svg issue 384

geokontop:

## Description
The problem is that when the network is created from a json object it keeps the network's sizes info in the attribute "sizes". While when created from the constructor the sizes attribute is null. In that case the sizes' info is kept in the attributes "inputSize", "outputSize" and "hiddenLayers". 

## Motivation and Context

## How Has This Been Tested?
Tested in the browser for networks created from the constructor and networks created with fromJSON(json) method

## Types of changes
 I had to add a line of code in  to-svg.js. 
if(!size) size = network.sizes;
So that both cases are checked.

## Author's Checklist:

## Reviewer's Checklist:


geokontop:
resolves issue 384

robertleeplummerjr:
Ty for the PR!  There seems to be a lot of cache files added with this PR.  Any chance of having them removed?

geokontop:
I don't know why that happened. It looks like babel's compilation recreates all the files, so they appear to be altered. I will repeat the procedure tomorrow with a new pr. Have a nice day,

robertleeplummerjr:
You rock Georgios.

Issue #412, opened 2019-07-06
How to run examples in the develop branch?


massaroni:
What command should we use to run an example in the develop branch?

## *What* is wrong?
Examples are throwing errors in the develop branch.

## *Where* does it happen?
In the develop branch.

## *How* do we replicate the issue?
When I try to run any example in node, it's throwing an error in the develop branch:

```bash
$ node examples/javascript/which-letter-simple.js 
.../brain.js/examples/javascript/which-letter-simple.js:1
(function (exports, require, module, __filename, __dirname) { import brain from '../../src';
                                                              ^^^^^^

SyntaxError: Unexpected token import
    at new Script (vm.js:51:7)
    at createScript (vm.js:136:10)
    at Object.runInThisContext (vm.js:197:10)
    at Module._compile (internal/modules/cjs/loader.js:618:28)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:665:10)
    at Module.load (internal/modules/cjs/loader.js:566:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:506:12)
    at Function.Module._load (internal/modules/cjs/loader.js:498:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:695:10)
    at startup (internal/bootstrap/node.js:201:19)


$ node --version
v9.11.2
```

## Expected behavior
It works fine in the master branch:

```bash
$ node examples/which-letter-simple.js 
iterations: 10, training error: 0.23317416631428975
iterations: 20, training error: 0.210598456317468
iterations: 30, training error: 0.17436649541030716
iterations: 40, training error: 0.13409059488488398
iterations: 50, training error: 0.10131207107612829
iterations: 60, training error: 0.07420014998226022
iterations: 70, training error: 0.0518410705540017
iterations: 80, training error: 0.03558307536025094
iterations: 90, training error: 0.025044735270968704
iterations: 100, training error: 0.018431227715397347
iterations: 110, training error: 0.014178778824250277
iterations: 120, training error: 0.011323259386754258
iterations: 130, training error: 0.009319155256412576
iterations: 140, training error: 0.007856361954302204
iterations: 150, training error: 0.0067525114866180585
iterations: 160, training error: 0.005895903210432278
iterations: 170, training error: 0.005215334626081061
a
```

mubaidr:
You have to change `import` (not supported in nodejs environment) statement to `require`, like: 
```js
const brain = require('../../src')
```

[...]

PR #413, opened 2019-07-08
run examples in nodejs without running a build

massaroni:
## Description
Run example scripts in nodejs without running a build.  This should save time for contributors, and make it easier for users to get comfortable when they first discover brain.js.

## Motivation and Context
This is a solution to [Issue #412](https://github.com/BrainJS/brain.js/issues/412) that enables maintainers and contributors to continue using their preferred es6 import statements, going forward.

For example, now you can run examples containing es6 imports, like this:
```bash
node -r esm examples/javascript/which-letter-simple.js
```

Overall, this PR has the following code changes: fixing some es6 exports for nodejs compatibility, added a new dev dependency, updated the dist build and readme documentation.


mubaidr:
Nice work! but I think a more general approach is needed, more info here: https://github.com/BrainJS/brain.js/issues/412#issuecomment-509223167

mubaidr:
Closing in favor of: https://github.com/BrainJS/brain.js/pull/415

PR #433, opened 2019-08-14
Update README.md

FinlayDaG33k:
image

## Description
I have added a line to the documentation that informs users that if they want to use the values (currently `error` and `iterations` of the `log`) property in their own output, they can do so by using the `callback` property instead.

## Motivation and Context

It's not a required change but I feel like it might save some hassle for new users :)

[issue](https://github.com/BrainJS/brain.js/issues/#431)

## Types of changes

- [x] Documentation update (<-- why is this not in the template?)

## Author's Checklist:

## Reviewer's Checklist:


robertleeplummerjr (approved):
Very nice job!  Ty!

Issue #446, opened 2019-09-16
Change the signature of logging function from log(info:string)  to log(info:string, {iterations: number, error: number})})


geosp:
## *What* is wrong?
I'm visualizing training by observing the model's behavior via a plot of iterations vs. error. 
I expected the log function signature to have iterations and error as numeric values. 

## *Where* does it happen?
It happens any time a training option for log is provided as a function.

## *How* do we replicate the issue?
1. Clone my repository [hello_brain](https://github.com/geosp/hello_brain)
2. Follow readme instructions.
3. Run `hpvsmpg` example like this: `npm start hpvsmpg`

## *How* important is this (1-5)?
3

## Expected behavior (i.e. solution)
I created a work around in my code in the [train helper](https://github.com/geosp/hello_brain/blob/master/src/core/train.ts#L59)
I have also created a PR associated with this issue.

## Other Comments
This is what I'm accomplishing with this feature:
![brain_prediction](https://user-images.githubusercontent.com/20133884/64932218-76bf5000-d80b-11e9-8a21-28e8e14a05bd.png)
![brain_error_log](https://user-images.githubusercontent.com/20133884/64932228-80e14e80-d80b-11e9-99f4-4b99b2d34cc9.png)

colus001:
I think you can use `trainingOptions.callback` instead. 

```
export interface INeuralNetworkTrainingCallback {
  (state: INeuralNetworkState): void;
}

export interface INeuralNetworkState {
  iterations: number;
  error: number;
}
```

I'm using like below.

```
const trainingOptions: INeuralNetworkTrainingOptions = {
  ...
  callback: (state) => progress.tick(callbackPeriod, {
    error: `error: ${state.error.toFixed(6)} / ${errorThresh}`,
  }),
  callbackPeriod,
};
```

geosp:
Thank you for your response. I wonder why we need both log and callback. I would be nice to just keep callback and get rid of log.

PR #450, opened 2019-09-16
Fix wrong log callback type

colus001:
Fix wrong log callback type on trainingOptions

cat
## Description

Only tiny fix to avoid confusion on log callback options

## Motivation and Context

[issue](https://github.com/BrainJS/brain.js/issues/446)
Kind of related

## How Has This Been Tested?

No test required

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/__tests__/neural-network/json.js
+++ b/__tests__/neural-network/json.js
@@ -457,7 +457,7 @@ describe('JSON', () => {
           const json = net.toJSON();
           const newNet = new NeuralNetwork()
             .fromJSON(json);
-          expect(newNet.trainOpts.log).toBe(console.log);
+          expect(newNet.trainOpts.log).toBe(net.logTrainingStatus);
         });
       });
 
--- a/__tests__/neural-network/json.js
+++ b/__tests__/neural-network/json.js
@@ -439,7 +439,7 @@ describe('JSON', () => {
             .fromJSON(json);
           expect(newNet.trainOpts.log).toBe(NeuralNetwork.trainDefaults.log);
         });
-        it('uses console.log for `true`', () => {
+        it('uses net.logTrainingStatus for `true`', () => {
           const net = new NeuralNetwork({ hiddenLayers: [2] });
           const log = true;
           net.trainingTick = () => {};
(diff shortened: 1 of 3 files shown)


mubaidr:
I think the trainingLogCallback and trianingCallback has same interface. Both returns iterations and error. So why a different type?

colus001:
No it doesn't. 

logCallback returns `string`
callback returns `state`

```
const trainingOptions: INeuralNetworkTrainingOptions = {
  ...
  callback: console.log,
  ...
}
```

Will print.

```
{ iterations: 40, error: 0.1505548765650579 }
{ iterations: 41, error: 0.07967034997254249 }
```

By the way, 

```
const trainingOptions: INeuralNetworkTrainingOptions = {
  ...
  log: console.log,
  ...
}
```

will print out as below returning `string` not `INeuralNetworkState`

```
iterations: 30, training error: 0.24721950121220654
iterations: 40, training error: 0.1505548765650579
```

I don't think logCallback is actually necessary but at least type should be fixed.
It took me a while to find what's wrong here, since type was both State

mubaidr (approved):

mubaidr:
Sounds good, but it would be even better if we can make arguments same for both. So that `state` is passed to user provided function not the string. 

Can you please update so that:
- If `log` is true, then use console log to print string message
- If `log` is function (provided by user) then pass state to the function

Changes required here: https://github.com/BrainJS/brain.js/blob/master/src/neural-network.js#L446

colus001:
@mubaidr I reverted the type changes, and fixed log related codes. any comment?

mubaidr (approved):

mubaidr:
Just  a minor step, there is a redundant interface now: : https://github.com/BrainJS/brain.js/blob/279befba202558f1e7e18af30408880e13404e1a/src/index.d.ts#L89

mubaidr:
Updated tests are failing too. Please have a look. 
https://travis-ci.org/BrainJS/brain.js/jobs/591940249#L277
https://travis-ci.org/BrainJS/brain.js/jobs/591940249#L296

mubaidr:
Everything except the above two changes required looks good, thanks for your patience and contribution. 👍

colus001:
```
$ npm run test ./__tests__/neural-network/json.js

> brain.js@2.0.0-alpha.9 test /Users/alma/Development/brain.js
> jest --env=node "./__tests__/neural-network/json.js"

 PASS  __tests__/neural-network/json.js

Test Suites: 1 passed, 1 total
Tests:       71 passed, 71 total
Snapshots:   0 total
Time:        0.698s, estimated 1s
Ran all test suites matching /.\/__tests__\/neural-network\/json.js/i.
```

`index.d.ts` no longer exists in this PR 😄

mubaidr (approved):

mubaidr:
Thanks again! Looking forward to more contributions!

PR #492, opened 2019-12-17
Fix NaN error rate bug in RNNs.

N8python:
I changed the NaN check along with the error rate calculation. I tried this locally on my computer and it causes the RNNs to be much, much less buggy.

This minor change prevents the RNNs from having an error rate of Infinity and NaN. I know why it prevents the NaN error rate (no more dividing by zero), however as I am new to ML I'm not sure why. I think it might have to do with `Number.isNaN` genuinely checking for NaN ( and not returning true for strings, arrays, etc.), but I don't know for sure.

This is a minor change, so it shouldn't have much of an impact on the rest of the library. I saw it's results while running some neural nets locally on my machine. I don't have an amazing testing environment, but the results looked promising.

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/src/recurrent/rnn.js
+++ b/src/recurrent/rnn.js
@@ -473,9 +473,9 @@ class RNN {
         const err = this.trainPattern(data[j], true);
         sum += err;
       }
-      error = sum / data.length;
+      error = sum / Math.max(data.length, 1);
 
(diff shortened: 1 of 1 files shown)


N8python:
Never mind. It was an exploding gradient.

robertleeplummerjr:
ty for making this, and updating the issue.

Issue #551, opened 2020-05-17
Migrate to TypeScript


mubaidr:
We plan to gradually migrate brain.js to TypeScript, code base is pretty large, so we would love your help! 💪

How to contribute?
- Convert a file from .js to .ts
- Add types, fix all type errors.
- Submit a PR! 🎉

**Here you can find a guide on [how to contribute](https://github.com/BrainJS/brain.js/blob/master/CONTRIBUTING.md).**

Want to convert something, let us know in the comment and go ahead! 😎

To avoid duplicate work please comment on which part you want to work on (as long as nobody else is working on it) so we can mark it as taken.

Reach out to us!
Feel free to reach if you have questions or need help getting started. You can leave comments here or you can tag me in your PR if you need any help or you're not sure about something!

You can also get in touch on our [![Gitter](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/brain-js/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) & [![Slack](https://slack.bri.im/badge.svg)](https://slack.bri.im).

Happy Coding! 🤟 

<hr/>

**UPDATE**: 

Wohoooo!!! 🎉 

All files inside `src` directory are migrated to typescript. (Except few ones which are already taken up and being worked on), though we are still looking on improvements to types in these files and removing `any` types from the source. You are welcome to contribute. 😊 

`__tests__` directory has still some files left that needs migration to typescript, so feel free to pick em up! 🍭 

<hr/>

**UPDATE**

For anyone looking to contribute, here is the list of files that still needs typescript migration: 

https://github.com/BrainJS/brain.js/search?l=javascript&p=1

yashshah1:
I'd like to start working on this, I have some experience with Ts, can you tell me which module is up for grabs?

mubaidr:
@yashshah1 You are welcome to contribute, please pick any module or portion for conversion and mention it here, so I can mark it in-progress (to avoid any duplication effort).

nabeelvalley:
I'd like to work on this too, any recommendations on where to start?

mubaidr:
@nabeelvalley here is the short guide, super simple to start. Please pick any module or portion for conversion and mention it here, so I can mark it in-progress (to avoid any duplication effort).

**Here you can find a guide on [how to contribute](https://github.com/BrainJS/brain.js/blob/master/CONTRIBUTING.md).**

nabeelvalley:
[...]

PR #587, opened 2020-08-30
Convert Utitlities/zeros* to typescript

harshkhandeparkar:
## Description
- Converted src/utilitities/zeros*.js to typescript
- To use typescript default imports, I either had to do require().default, change the default exports to just exports (*or convert the whole codebase to typescript*)
So I chose option 2 since it is better to [avoid default exports](https://basarat.gitbook.io/typescript/main-1/defaultisbad) anyway.
- Changed all the files that import the zeros* utilities.

NOTE: This branch is based on the `ts` branch so merging #585 is required.

## Motivation and Context

## Types of changes

- [x] ¯\\\_(ツ)_/¯

## Author's Checklist:

## Reviewer's Checklist:


src/estimator/mean-squared-error.ts
```diff
@@ -0,0 +1,77 @@
...
+import { IKernelRunShortcut } from 'gpu.js';
+
+/**
+ * 2D Mean Squared Error
+ */
+export function mse2d(
+  this: {
+    constants: { height: number; width: number; length: number };
+  },
+  errors: Array<[number, number]>
```
harshkhandeparkar, line 11: Is this the linter? This should have been in the older commits. I believe the linter is running after the commit. Is this a known issue?

mubaidr:
Please pull `master` into `HarshKhandeparkar:js-needs-types`

src/estimator/mean-squared-error.ts
mubaidr: Please pull master into HarshKhandeparkar:js-needs-types

harshkhandeparkar:
Okay.

src/estimator/mean-squared-error.ts
harshkhandeparkar: Done. This is still there though. Should it happen precommit?

mubaidr: We run `npm run lint` as `pre-commit` hook. So if there is any error/warning in code it will be notified.

mubaidr (approved):

src/estimator/mean-squared-error.ts
harshkhandeparkar: Ohk. I used the VSCode git extension so I didnt get any errors.
If that hook makes any changes, doesn't it commit those changes?

PR #589, opened 2020-09-01
Convert utilities/values* to typescript

harshkhandeparkar:
## Description
- Converted utilities/values* to ts
- Apparently it isnt used anywhere (¯\\\_(ツ)_/¯)so this part was easy 

## Motivation and Context

## Types of changes

- [x] ¯\\\_(ツ)_/¯

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/src/utilities/values-2d.ts
+++ b/src/utilities/values-2d.ts
@@ -8,7 +8,7 @@ export function values2D(
   height: number,
   value: number
 ): Float32Array[] {
-  const result = new Array(height);
+  const result: Float32Array[] = new Array(height);
   for (let y = 0; y < height; y++) {
     result[y] = values(width, value);
   }
--- a/src/utilities/values-2d.ts
+++ b/src/utilities/values-2d.ts
@@ -1,0 +1,16 @@
+import { values } from './values';
+
+/**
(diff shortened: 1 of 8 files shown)


src/utilities/values-2d.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Returns a matrix of given width and height with each element filled with the same value
+ */
+export function values2D(
+  width: number,
+  height: number,
+  value: number
+): Float32Array[] {
+  const result = new Array(height);
```
mubaidr: I think `Float32Array` constructor should be preffered here.

harshkhandeparkar: I dont that will work because Float32Arrays need to have Float32s inside, not other Float32Arrays.

mubaidr: return type is `Float32Array[]` and we are returning `result` here which is `new Array(height)`

harshkhandeparkar: Ohk. I'll make it result: `Float32Array[]` then.

harshkhandeparkar: Done ig.

PR #590, opened 2020-09-01
Refactor utilities/cast to Typescript

nabeelvalley:
Update the `utilities/cast.js` file to use typescript -> `utilities/cast.ts`

## Description

The change does the following:

- Convert `module.exports` to an `export` on each function
- Adds type annotations to functions and 

## Motivation and Context

Partially addresses #551 

## How Has This Been Tested?

Only additions of type annotations to existing code, issue with tests was prevalent before making changes

Typescript check is fine without any issues

## Types of changes

## Author's Checklist:

All existing logic has been retained, just the type annotations added

## Reviewer's Checklist:

### DIFF
--- a/src/utilities/cast.ts
+++ b/src/utilities/cast.ts
@@ -23,7 +23,7 @@ export function objectsToFloat32Arrays(
   table: Record<string, number>,
   length: number
 ): Float32Array[] {
-  const results = [];
+  const results: Float32Array[] = [];
   for (let i = 0; i < objects.length; i++) {
     const object = objects[i];
     const result = new Float32Array(length);
--- a/src/utilities/cast.ts
+++ b/src/utilities/cast.ts
@@ -1,21 +1,28 @@
-function arraysToFloat32Arrays(arrays) {
-  const result = [];
+export function arraysToFloat32Arrays(arrays: number[][]): Float32Array[] {
+  const result: Float32Array[] = [];
   for (let i = 0; i < arrays.length; i++) {
     result.push(Float32Array.from(arrays[i]));
   }
   return result;
 }
-function arrayToFloat32Arrays(array) {
-  const result = [];
+
+export function arrayToFloat32Arrays(array: number[]): Float32Array[] {
+  const result: Float32Array[] = [];
   for (let i = 0; i < array.length; i++) {
     result.push(Float32Array.from([array[i]]));
   }
   return result;
 }
-function arrayToFloat32Array(array) {
+
+export function arrayToFloat32Array(array: number[]): Float32Array {
   return Float32Array.from(array);
 }
-function objectsToFloat32Arrays(objects, table, length) {
+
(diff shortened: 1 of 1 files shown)


src/utilities/cast.ts
```diff
@@ -1,21 +1,28 @@
-function arraysToFloat32Arrays(arrays) {
-  const result = [];
+export function arraysToFloat32Arrays(arrays: number[][]): Float32Array[] {
```
mubaidr, line 1: Since we have updated export style. i.e. from default to named exports, I think we need to update the referred imports too.  (References for these exports).

src/utilities/cast.ts
```diff
@@ -1,21 +1,28 @@
...
   return Float32Array.from(array);
 }
-function objectsToFloat32Arrays(objects, table, length) {
+
+export function objectsToFloat32Arrays(
+  objects: Array<Record<string, number>>,
+  table: Record<string, number>,
+  length: number
+): Float32Array[] {
   const results = [];
```
mubaidr: I think `Float32Array` constructor should be preferred here.

mubaidr:
Awesome job! 👍 Just few minor things and we are good to go.

src/utilities/cast.ts
nabeelvalley: The `result` variable here is a an array of `Float32Array`s,  I can make that more apparent by setting the type as:

```js
const results: Float32Array[] = []
```

Does this suggestion seem correct?

src/utilities/cast.ts
nabeelvalley: The files that use these utils are 

`recurrent/rnn-time-step.js`

```js
const {
  arraysToFloat32Arrays,
  arrayToFloat32Arrays,
  objectsToFloat32Arrays,
  objectToFloat32Arrays,
  objectToFloat32Array,
} = require('../utilities/cast');
```

`neural-network.js`:

```js
const { arrayToFloat32Array } = require('./utilities/cast');
```

Which are using the restructured imports, is that correct? If not, please advise

nabeelvalley:
Thanks @mubaidr for the response, I'll make the changes once you've confirmed based the comments above which of the changes I should make

src/utilities/cast.ts
mubaidr: Yes, we can make it apparent by using `: Float32Array[]`.

src/utilities/cast.ts
mubaidr: Yes, this is correct. 👍

src/utilities/cast.ts
nabeelvalley: Fixed this in the latest commit

mubaidr:
Looks all good. We are trying to fix config for build scripts, once it is done, we will merge this. 👍

mubaidr:
Thank you! 💯

PR #591, opened 2020-09-03
refactor: convert utilities/kernel to ts

harshkhandeparkar:
## Description
- Converted utilities/kernel to typescript
- Updated estimator/mean-squared-error because of this change
- ESLint doesn't like non-null-assertion-operator `!` but due to the way it has been set up, I couldn't avoid it. (ESLint does throw a warning though)

## Motivation and Context

## Types of changes

- [x] ¯\\\_(ツ)_/¯

## Author's Checklist:

## Reviewer's Checklist:


robertleeplummerjr (approved):
LGTM!  TY!

PR #593, opened 2020-09-09
Convert max.js and to-array.js to typescript

nabeelvalley:
- Convert the `utilities/max` and `utilities/to-array` to Typescript
- Replaces referenced default imports with named imports

## Motivation and Context

Contributes to #551 

## How Has This Been Tested?

Ran the unit tests, verified the results for tests where these functions are used (it appears some of the other tests are still broken)

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/__tests__/utilities/max.js
+++ b/__tests__/utilities/max.js
@@ -1,4 +1,4 @@
-const max = require('../../src/utilities/max');
+const { max } = require('../../src/utilities/max');
 
 describe('max', () => {
   test('should find max in object', () => {
--- a/src/utilities/max.ts
+++ b/src/utilities/max.ts
@@ -1,4 +1,5 @@
-const { toArray } = require('./to-array');
+import { toArray } from './to-array';
+
 /**
  *
  * @param values
--- a/src/utilities/max.ts
+++ b/src/utilities/max.ts
@@ -1,0 +1,9 @@
+const { toArray } = require('./to-array');
+/**
+ *
+ * @param values
+ * @returns {number}
+ */
+export function max(values: Record<string, number> | number[]): number {
+  return Math.max(...toArray(values));
+}
--- a/src/utilities/to-array.ts
+++ b/src/utilities/to-array.ts
@@ -3,9 +3,11 @@
  * @param values
  * @returns {*}
  */
-module.exports = function toArray(values) {
+export function toArray(
+  values: Record<string, number> | number[]
+): number[] | Float32Array {
   if (Array.isArray(values)) {
     return values;
   }
   return new Float32Array(Object.values(values));
-};
+}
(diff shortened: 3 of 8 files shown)


__tests__/utilities/max.js
```diff
@@ -1,4 +1,4 @@
-const max = require('../../src/utilities/max');
+const { max } = require('../../src/utilities/max');
```
mubaidr, line 1: I think we should use imports instead of require.

src/utilities/max.ts
```diff
@@ -0,0 +1,9 @@
+const { toArray } = require('./to-array');
+/**
+ *
+ * @param values
+ * @returns {number}
+ */
+export function max(values: Record<string, number> | number[]): number {
```
mubaidr, line 8: I think `values` should be `number|number[]` here, since we are not dealing with strings here.

src/utilities/to-array.ts
```diff
@@ -3,9 +3,11 @@
  * @param values
  * @returns {*}
  */
-module.exports = function toArray(values) {
+export function toArray(
+  values: Record<string, number> | number[]
```
mubaidr, line 7: Again, type should be `number | number[]`

mubaidr: Well, I think this is all good here. `Record<string, number> | number[]` is fine.

src/utilities/max.ts
nabeelvalley: @mubaidr the `string` refers to the key, the `number` is the value - this is due to the following unit test. The `string` refers to the object's keys (`a`, `b`, `c`, `d`) 

`__tests__/utilities/to-array.js`

```js
const { toArray } = require('../../src/utilities/to-array');

describe('to-array', () => {
  test('should convert object to array', () => {
    const obj = { a: 1, b: 5, c: 10, d: 0 };

    expect(toArray(obj)).toBeInstanceOf(Float32Array);
    expect(toArray(obj).length).toBe(4);
  });
});
```

mubaidr: Oh ok. I missed that.

__tests__/utilities/max.js
nabeelvalley: Which files would you like me to change? Just the one?

Changing some (such as `neural-networks.js`) requires a lot of files in the codebase to instead use the `.default`

mubaidr: I have updated imports except in `__tests__` which is not require at the time.

mubaidr:
Thank you! 👍

nabeelvalley:
@mubaidr thanks for making those changes, as a general question, should we be just importing with `import` for where we're referencing a ts file and leave other imports as-is (with`require`)?

mubaidr:
Yes, that is the plan o prefer named exports in typescript modules. As we migrate a module to typescript, we should update related `require`s to `import`s for that module.

PR #599, opened 2020-09-17
Convert 2 files to Typescript

ofirgeller:
Converts the array-lookup-table and data-formater files to Typescript 

## Description
Besides notations and comments I made the following changes:

Used a class instead of a function for  'array-lookup-table'

countSpecial had 2 bugs and is not called from anywhere, since it is not 100% clear what it was meant to do so I deleted it.
Bugs: 
Checking if i is less than the array instead of the length of the array
Setting the startIndex of the next search to a boolean, this looks like an attempt at an optimization (start search from the last found item) but it is incorrect.

If anyone can say for sure what the method should do (return the number of special indexes that exist in the output parameter"?) I can re-write it.

added a failing branch to catch unexpected "values" values, once the callers are converted it can be removed.

## How Has This Been Tested?
Tests were broken before the change, I did run lint:typecheck and it seems ok.

## Reviewer's Checklist:


ofirgeller:
@robertleeplummerjr any ETA for the review? I'm ready to move on to other files but I want to see we are on the same page first.

src/utilities/data-formatter.ts
```diff
@@ -291,4 +300,4 @@ function defaultRNNFormatter(data) {
   return result;
 }
 
-module.exports = { DataFormatter, defaultRNNFormatter };
+export default DataFormatter;
```
robertleeplummerjr: We've not been using default exports.  Can we just use named ones?

src/utilities/data-formatter.ts
```diff
@@ -1,54 +1,65 @@
...
+    const tempCharactersTable: any = {};
     for (
       let dataFormatterIndex = 0, dataFormatterLength = values.length;
       dataFormatterIndex < dataFormatterLength;
       dataFormatterIndex++
     ) {
       const characters = values[dataFormatterIndex];
 
-      if (characters.hasOwnProperty('length')) {
+      if ((characters as any).hasOwnProperty('length')) {
```
robertleeplummerjr, line 36: This `any` is a bit concerning.

src/utilities/data-formatter.ts
```diff
@@ -85,7 +96,7 @@ class DataFormatter {
     return result;
   }
 
-  toIndexesInputOutput(value1, value2 = null, maxThreshold = 0) {
+  toIndexesInputOutput(value1: string | number | string[], value2: string | string[] | null = null, maxThreshold = 0) {
```
robertleeplummerjr: Perhaps this is `Values`?

src/utilities/data-formatter.ts
```diff
@@ -1,54 +1,65 @@
...
-    this.characters = [];
-    this.specialIndexes = [];
+
     this.buildCharactersFromIterable(values);
     this.buildTables(maxThreshold);
   }
 
-  buildCharactersFromIterable(values) {
-    const tempCharactersTable = {};
+  buildCharactersFromIterable(values: Values) {
```
robertleeplummerjr: Each method needs a return type.

src/utilities/data-formatter.ts
```diff
@@ -148,27 +159,27 @@ class DataFormatter {
...
-  static fromAllPrintable(maxThreshold, values = ['\n']) {
+  static fromAllPrintable(maxThreshold: number, values = ['\n']) {
     for (let i = 32; i <= 126; i++) {
       values.push(String.fromCharCode(i));
     }
     return new DataFormatter(values, maxThreshold);
   }
 
-  static fromAllPrintableInputOutput(maxThreshold, values = ['\n']) {
+  static fromAllPrintableInputOutput(maxThreshold: number, values = ['\n']) {
```
robertleeplummerjr: `values` is missing a type

src/utilities/data-formatter.ts
```diff
@@ -223,32 +226,38 @@ var dataFormatter = {
...
+ */
+type DefaultRNNFormatterThis = { dataFormatter?: DataFormatter, formatDataIn: Function };
+
 /**
  *
  * @param {*[]} data
  * @returns {Number[]}
  */
-function defaultRNNFormatter(data) {
+export function defaultRNNFormatter(this: DefaultRNNFormatterThis, data: any[]): number[] {
```
robertleeplummerjr: Do we need a `this` here?  Since this is non-extendable, we can enforce `data` type, which is `string` here.

robertleeplummerjr (changes requested):
A few things.  This also breaks the RNN classes, so we'll want to put some attention on that before closing this into master.  @mubaidr were you wanting this to merge directly into master?  If so, we'll be blocked for a bit to do a release.

src/utilities/data-formatter.ts
ofirgeller: an import in  src/index.ts:
`import DataFormatter from './utilities/data-formatter';`
is assuming there is a default export, or it needs to be converted to 
`import * as DataFormatter from './utilities/data-formatter'; `

The naming makes me think the first option is what was intended, but It is a change, should I change it to the second option?

src/utilities/data-formatter.ts
ofirgeller: "This any is a bit concerning."

We don't need the any, but I put it there as a warning that the type is not what we think it is.

Checking for the existence of the 'length' property does not help typescript narrow the type, so some casting is needed (next line after the check).

The reason why characters has the wrong type is that I typed 'Values' as 
`type Values = string[]  | number[] | string;`
But I confirmed that during runtime it can actually also be string[][], so:

`type Values = string[] | string[][] | number[] | string;`

But if I add this option it causes issues with this line:

`const character = values[dataFormatterIndex];`
and then
` this.characters.push(character);`

These lines looks a bit strange, since they are re-indexing into values instead of using the 'characters' variable which we now know is a number.

I did not want to change these lines in this commit (and before we get the tests passing again)

If you prefer I can add the nested array option to 'Values' , and then change the code so we are not re-indexing into values, it sounds like a good idea to me.

src/utilities/data-formatter.ts
ofirgeller: Probably , It depends on the likely change to the values type mentioned in the previous comment, the code might not be able to handle "string[][]" correctly so I left this for later.

src/utilities/data-formatter.ts
ofirgeller: It has the return type of "void" (will actually return undefined') automatically inferred by Typescript,  I can make it explicit:
` buildCharactersFromIterable(values: Values): void {`

If that is what we are doing everywhere.

src/utilities/data-formatter.ts
ofirgeller: TS infers it based on the default parameter value to be string[] which I think is the correct type.

src/utilities/data-formatter.ts
ofirgeller: 'data' is not a string, or at least that is not what the function body is treating it as, for exemple:
` Array.isArray(data[0]`

The function is referencing 'this' in a few places, so we need to type it, not sure what you mean by 'non-extendable'. this function is used as a "method" since it is set to an object property and then invoked as a property of the object (which sets the 'this' keyword to the object) and in testing it is invoked using 'call' to set 'this' to a mock object.

src/utilities/data-formatter.ts
robertleeplummerjr: My IDE asks me to do this, I assumed it was part of the rules we have setup.

src/utilities/data-formatter.ts
robertleeplummerjr: We're trying to conform and be as verbose as possible.  I think my preference would be to imply as little as possible if we're going to fully adopt typescript.

src/utilities/data-formatter.ts
robertleeplummerjr: Ah, ty.

src/utilities/data-formatter.ts
robertleeplummerjr: Go ahead and use named exports, and change wherever there are imports.

ofirgeller: Done

src/utilities/data-formatter.ts
ofirgeller: keeping it as is for now since values now includes string[][] and this function can't handle that.

src/utilities/data-formatter.ts
ofirgeller: Done

src/utilities/data-formatter.ts
ofirgeller: Done for all methods/functions in the data-formatter file

ofirgeller:
@robertleeplummerjr Please take a look, if you want me to rebase into a single commit or anything else let me know.

robertleeplummerjr (approved):
LGTM.  Ty for your hard work.

PR #605, opened 2020-09-30
Lookup ts

robertleeplummerjr:

## Motivation and Context

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


mubaidr (approved):

src/lookup.ts
```diff
@@ -168,61 +179,64 @@ class Lookup {
...
-    shape.push(typeof data);
     return shape;
-  }
+  },
 
-  static addKeys(value, table) {
-    if (Array.isArray(value)) return;
+  addKeys(value: number[] | INumberHash, table: INumberHash): INumberHash {
+    if (Array.isArray(value)) return table;
     table = table || {};
```
mubaidr: We might not need this line, as `table` is required and can't be null. so no need to re-assign.

mubaidr (approved):

src/lookup.ts
robertleeplummerjr: Ty sir.  And done.

PR #607, opened 2020-10-04
convert reccurrent/matrix to ts

mubaidr:

## Motivation and Context

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


src/recurrent/matrix/index.ts
```diff
@@ -21,64 +23,83 @@ class Matrix {
    *
    * @param {Number} row
    * @param {Number} col
-   * @returns {Float32Array|Array}
+   * @returns {Number}
```
mubaidr, line 27: @robertleeplummerjr Can you please check this piece?

src/recurrent/matrix/equation.ts
```diff
@@ -157,94 +194,114 @@ class Equation {
...
-      left: m,
-      get right() {
-        return self.inputRow;
-      },
       product,
-      forwardFn: rowPluck,
-      backpropagationFn: rowPluckB,
+      left: matrix,
+      // TOFIX: returns `number` but expected `Matrix`
+      // get right() {
```
mubaidr, line 225: @robertleeplummerjr need help here, please

src/recurrent/matrix/index.ts
robertleeplummerjr: I believe this is an unused method that was aimed at mirroring https://github.com/karpathy/recurrentjs/blob/84d69349c5412ddebd2b1cbc9b29861ddf3a8169/src/recurrent.js#L63

Since this should actually be "getWeight", non-plural.

src/recurrent/matrix/equation.ts
robertleeplummerjr: For the time being, to keep it as close to the original as possible, I'd just have an `ts-ignore` there.  The new networks that allow for network composition are replacing these.

mubaidr: Ok Cool.

PR #608, opened 2020-10-04
feat: Convert feed-forward.js to feed-forward.ts and tests

robertleeplummerjr:
feat: Convert lookup-table.js to lookup-table.ts
feat: Convert momentum-root-mean-squared-propagation.js to ts
fix: Various fixes to make layers and praxises JSON to work together
fix: Change JSON "name" property, to both type and title

## Motivation and Context

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


src/layer/base-layer.ts
```diff
@@ -30,7 +50,7 @@ export interface ILayerSettings {
   depth?: number | null;
   weights?: KernelOutput | null;
   deltas?: KernelOutput | null;
-  name?: string | null;
+  title?: string | null;
```
mubaidr, line 53: We are dealing here with both `undefined and null`, not a big deal but I think `?` should be preferred.

src/praxis/momentum-root-mean-squared-propagation.ts
```diff
@@ -0,0 +1,175 @@
...
+      Math.sqrt(momentum + this.constants.smoothEps) -
+    this.constants.regularizationStrength * weight
+  );
+}
+
+export function isClippedByValue(
+  value: number,
+  max: number,
+  min: number
+): number {
```
mubaidr, line 72: we can reverse the logic to simplify code and also use `boolean` matching the function naming.

mubaidr (approved):

src/layer/base-layer.ts
mubaidr: nevermind, we can update it latter if required.

PR #611, opened 2020-10-05
convert `layer-from-json.js` to typescript

urwrstkn8mare:
Convert `src\utilities\layer-from-json.js` to `src\utilities\layer-from-json.ts`. I also had to make a minor change to `src\layer\index.ts`.

## Description
My goal was to convert `src\utilities\layer-from-json.js` to `src\utilities\layer-from-json.ts` (https://github.com/BrainJS/brain.js/issues/551#issuecomment-703368824). I used `ILayerJSON` from https://github.com/BrainJS/brain.js/pull/608 and that's why I'm submitting a PR to [`feed-forward-ts`](https://github.com/BrainJS/brain.js/tree/feed-forward-ts). 

I hope this is alright but I was a little unsure about things so I'd appreciate any feedback if this PR isn't up to standards.

## Motivation and Context

https://github.com/BrainJS/brain.js/issues/551

## How Has This Been Tested?

I have run `npm run test` both before and after my changes and I don't think there is much difference in the result. `npm run lint` shows some warnings but non-related to the files I have worked on as far as I can see. 

In summary, I think it would be good if someone else made sure I haven't broken anything as it is really hard to tell with `npm run test` as there is a lot of preexisting failures. 

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


src/layer/index.ts
```diff
@@ -33,7 +33,11 @@ export { Target, target } from './target';
 export { Transpose, transpose } from './transpose';
 export { Zeros, zeros } from './zeros';
 
-export const layerTypes = {
+interface layerTypesMap {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
```
mubaidr, line 38: There should be a way to avoid `any` type

src/utilities/layer-from-json.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable prettier/prettier */
+import { layerTypes } from '../layer';
+import { ILayer, ILayerJSON } from '../layer/base-layer';
+
+export default function layerFromJSON(
+  jsonLayer: ILayerJSON,
+  ...args: unknown[]
```
mubaidr, line 7: I think args here refer to as `ILayerSettings`. More info: https://github.com/BrainJS/brain.js/blob/master/src/layer/base-layer.ts#L27

src/utilities/layer-from-json.ts
```diff
@@ -0,0 +1,17 @@
...
+): ILayer {
+  const Layer = layerTypes[jsonLayer.type];
+
+  const realLayer = new Layer(...args);
```
mubaidr, line 11: Again, `Layer` has `any` type :( We might run into issue if layer type is `Internal` (which is static)

mubaidr (changes requested):
Thanks for the contributions. 

This one has some tricky parts, looking forward to more commits.

src/utilities/layer-from-json.ts
urwrstkn8mare: Yeah, I knew that would come up :). Sorry, but I'm a little confused about how I'm supposed to handle it. I know that at its core, the layer type can be an `ILayer` or `InternalModel`. Some types have constructors, others don't. So how do I construct layers that don't have constructors?

src/utilities/layer-from-json.ts
urwrstkn8mare: Sorry - correct me if I'm wrong - but doesn't some of the layers have `ILayerSettings` and `ILayer` as args?

Eg. (from `src\layer\activation.ts`)
```typescript
  constructor(inputLayer: ILayer, settings?: ILayerSettings) {
```

src/layer/index.ts
urwrstkn8mare: I think I'll deal with https://github.com/BrainJS/brain.js/pull/611#discussion_r499437605 first and then this should be pretty easy.

src/utilities/layer-from-json.ts
mubaidr: I think we might need to have an `enum` collection for all layer types. (already available in `layer.type` property). Then we can call appropriate constructor based on the comparison.

src/utilities/layer-from-json.ts
urwrstkn8mare: I think also `inputLayer` is supplied by here: https://github.com/BrainJS/brain.js/blob/f572a3d2489950cf530d7f7cee7721167744d351/src/feed-forward.ts#L697

In `layerFromJSON()`'s args.

mubaidr: Lets focus on getting types for layer constructor, then this will be easier to handle. https://github.com/BrainJS/brain.js/pull/611#discussion_r499871647

src/utilities/layer-from-json.ts
urwrstkn8mare: Sorry I'm not very familiar with this project, but would be the appropriate constructors?

robertleeplummerjr: >Some types have constructors, others don't. 

That shouldn't be the case.  Even if a layer is compound (ie, consisting of several internal layers) they are all translated to several layers.  An example is from: https://github.com/BrainJS/brain.js/blob/f572a3d2489950cf530d7f7cee7721167744d351/src/layer/feed-forward.ts#L9

feed-forward layer is translated to:
* Random
* Random
* Multiply
* Add
* Sigmoid

urwrstkn8mare: Oh, I was looking at layers such as `Internal` and `InternalModel`.

urwrstkn8mare:
@mubaidr and @robertleeplummerjr, thanks a lot for your time but I think I may be in over my head especially since this is my first look at this project (thus I'm not very familiar with it). Is it ok if you continue with this (I've allowed for edits by maintainers) or make a new PR?

mubaidr:
Don't worry we will fix this issue together.

mubaidr:
@robertleeplummerjr Can you please review these pull requests. https://github.com/BrainJS/brain.js/pull/626 & https://github.com/BrainJS/brain.js/pull/611

mubaidr:
https://github.com/BrainJS/brain.js/pull/611

PR #626, opened 2020-10-12
Add types to layer-from-json

Nelias:

## Description

I changed the `layer-from-json.js` file to a TypeScript file. This change required a different access of a property from both layer and `jsonLayer` object so I used a `get()` method from the lodash library (I had to add lodash types in package.json to use it). I did not change the structure of a function and this enabled its use in feed-forward.ts file properly. There is `any` used once as `Reflect.construct()` returns this type. In feed-forward.ts `getLayer()` is still not prepared to receive a `null` value returned from the `layerFromJSON()` function and in my opinion it is up to the maintainers what to do with this case.

## Motivation and Context

[Migrate to TypeScript #551](https://github.com/BrainJS/brain.js/issues/551)
[#611 allowed new PRs](https://github.com/BrainJS/brain.js/pull/611#issuecomment-704680150)

## How Has This Been Tested?

I ran `npm run test` before and after my changes. 13 test failed on Windows before and after my changes and in the files not connected to this function.

The most common fail was of this type:
`deltasRows = deltasRows ?? weightRows;
                                     ^
    SyntaxError: Unexpected token '?'`

I tried to add a unit test for this function but jest.config is still not properly set for all types of .ts files or even plain .js files that import .ts modules. Here is an error: 

broken-jest-config-for-typescript

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


src/utilities/layer-from-json.ts
```diff
@@ -0,0 +1,22 @@
...
+import { get } from 'lodash'
+import { ILayer, ILayerJSON } from '../layer/base-layer';
+
+export default function layerFromJSON(jsonLayer: ILayerJSON, inputLayer1?: ILayer) { 
+  if (!layer.hasOwnProperty(jsonLayer.type)) return null;
+
+  const Layer: Function = get(layer, `${layer}.${jsonLayer.type}`);
+
+  // eslint-disable-next-line
+  const realLayer: ILayer | any = Reflect.construct(Layer, arguments)
```
mubaidr: Can we not cast result for construct to `ILayer`?

mubaidr:
Awesome work! 👍🏼 

But I don't think adding a whole new library is worth for this small utility, what do you think? We can try to implement this method by own or use the `lodash.get` if it is not possible to write it our-self.

src/utilities/layer-from-json.ts
```diff
@@ -0,0 +1,22 @@
+import * as layer from '../layer'
+import { get } from 'lodash'
+import { ILayer, ILayerJSON } from '../layer/base-layer';
+
+export default function layerFromJSON(jsonLayer: ILayerJSON, inputLayer1?: ILayer) {
```
mubaidr, line 11: we have an extra unused parameter `inputLayer1` here.

src/utilities/layer-from-json.ts
```diff
@@ -0,0 +1,22 @@
+import * as layer from '../layer'
```
mubaidr: Missing semi's here, I think you forgot to run `npm run lint`

mubaidr (changes requested):

mubaidr:
As for you unit test, error is due to `??` operator, as the current babel config does not support it. I am updating it to include support for this. You can go ahead and add test, please. 👍🏼

src/utilities/layer-from-json.ts
Nelias: This is here because when this function is used in feed-forward.ts it requires two arguments when implemented in typescript, it did not work before in JS so I have no idea why it was there in the first place. Now it is here to not to make changes to the feed-forward.ts file.

src/utilities/layer-from-json.ts
Nelias: That is right

src/utilities/layer-from-json.ts
Nelias: We can but then `feed-forward.ts` breaks due to the possible early returned null from layerFromJSON.

Nelias:
@mubaidr I checked in the package.json and lodash was already present in the repository. I just added `@types/lodash`, which you would require anyway when moving to TypeScript in the whole project.

mubaidr:
Strange, we are not using `lodash` in `brainjs` not even in two of the dependency packages. 
you  can check dependencies here: https://www.npmjs.com/package/brain.js

Nelias:
Edit: in package-lock.json, it was probably a dependency of a dependency, not a direct dependency in package.json : )

Nelias:
truth

package.json
```diff
@@ -29,7 +29,8 @@
     "@types/babel__preset-env": "7.9.1",
     "@types/eslint": "^7.2.4",
     "@types/eslint-plugin-prettier": "^3.1.0",
-    "@types/prettier": "^2.1.2",
+    "@types/lodash": "^4.14.162",
```
mubaidr: Can we get rid of `lodash` please?

__tests__/utilities/layer-from-json.ts
```diff
@@ -0,0 +1,49 @@
...
+    const mockedLayer = mockLayer({
+      width: 3,
+      height: 2,
+      weights: null,
+      deltas: null,
+    });
+
+    expect(layerFromJSON(jsonLayer)).toEqual(mockedLayer);
+  });
+});
```
mubaidr: missing new line at the end of file :p

src/utilities/layer-from-json.ts
```diff
@@ -0,0 +1,22 @@
+import * as layer from '../layer';
+import { get } from 'lodash';
+import { ILayer, ILayerJSON } from '../layer/base-layer';
+
+export default function layerFromJSON(jsonLayer: ILayerJSON, inputLayer1?: ILayer) { 
+  if (!layer.hasOwnProperty(jsonLayer.type)) return null;
+
+  const Layer: Function = get(layer, `${layer}.${jsonLayer.type}`);
```
mubaidr: `Function` should be replaced by proper `Layer` interface/type. Instead of get we have to use own method to get later. For more information please see the conversation on here: https://github.com/BrainJS/brain.js/pull/611/files

mubaidr:
@robertleeplummerjr please

package.json
Nelias: Okay, I will try to find a different solution for that function. Maybe at some point in the future lodash will not be anymore in the deps of deps, so getting rid of it is a good idea after all.

src/utilities/layer-from-json.ts
Nelias: Reflect.construct() does not accept anything else than Function :)

__tests__/utilities/layer-from-json.ts
Nelias: done

package.json
Nelias: done

PR #630, opened 2020-10-13
Change all remaining utilities tests and three main tests to TypeScript

Nelias:

## Description

Seven files in total have been changed to TypeScript, some of them required changes in the old test logic.

## Motivation and Context

[Migrate to TypeScript #551](https://github.com/BrainJS/brain.js/issues/551)

## How Has This Been Tested?

I ran `npm run test` before and after my changes.

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


__tests__/istanbul-linker-util.ts
```diff
@@ -1,18 +1,24 @@
...
 class Recurse {
-  constructor(settings) {
+  settings: TSettings;
+
+  constructor(settings: TSettings) {
     this.settings = settings;
   }
 
-  into(ast) {
+  into(ast: Node | any) {
```
mubaidr: I this `ast` should be optional here with type `Node`.

__tests__/istanbul-linker-util.ts
```diff
@@ -136,14 +142,14 @@ class Recurse {
   }
 }
 
-module.exports = function (fn) {
+module.exports = function (fn: Object) {
```
mubaidr: `Object` is a generic, we may want a `function` definition here.

mubaidr (changes requested):

__tests__/istanbul-linker-util.ts
Nelias: There are no properties like 'type' existing on Node. In general istanbul is not even a typed library so this file should not be converted to TypeScript at all but if it is then `any` is what we have to use.

__tests__/istanbul-linker-util.ts
Nelias: done

PR #642, opened 2020-10-21
Update jest config file

eliamartani:
## Description
This PR fixes one issue I found when running `npm test`.

## Motivation and Context

For testing, the `??` is not recognized by Jest, setting `babelConfig` helps running `npm test`.

## How Has This Been Tested?

I retested the script after implementing it and the issue was gone.

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -4,9 +4,11 @@
   "coverageDirectory": "__coverage__",
   "globalSetup": "",
   "globals": {
-    "ts-jest": {}
+    "ts-jest": {
+      "tsConfig": "tsconfig.test.json"
+    }
   },
-  "preset": "ts-jest/presets/js-with-babel",
+  "preset": "ts-jest/presets/js-with-ts",
   "setupFiles": [],
   "testEnvironment": "node",
   "testPathIgnorePatterns": [
--- a/tsconfig.test.json
+++ b/tsconfig.test.json
@@ -1,0 +1,9 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "esModuleInterop": true,
+    "module": "es6",
+    "moduleResolution": "node",
+    "target": "es6"
+  }
+}


package.json
```diff
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "gpu.js": "^2.10.4",
-    "thaw.js": "^2.1.0"
+    "thaw": "^2.0.1"
```
mubaidr: We are using this package: 
https://www.npmjs.com/package/thaw.js

mubaidr (changes requested):

package.json
eliamartani: You're right, I messed up the plugins

mubaidr:
This sure does fix the issue. 👍🏼 

But what if we override `tsconfig` for tests to replace with `"target": "esnext"` with `"target": "es6"`, this will make sure typescript compiles newest feature to `es6` which `jest` understands. 

So instead of `Babel -> Typescript > jest` we can do `typescript > jest`

eliamartani:
> This sure does fix the issue. 👍🏼
> 
> But what if we override `tsconfig` for tests to replace with `"target": "esnext"` with `"target": "es6"`, this will make sure typescript compiles newest feature to `es6` which `jest` understands.
> 
> So instead of `Babel -> Typescript > jest` we can do `typescript > jest`

It works as well. `ts-jest` will receive `tsConfig`* instead with the configuration to be used, `tsconfig.test.json`

```json
{
  "tsConfig": "tsconfig.test.json"
}
```

tsconfig.test.json:
```json
{
  "compilerOptions": {
    "allowJs": true,
    "esModuleInterop": true,
    "module": "es6",
    "moduleResolution": "node",
    "target": "es6"
  }
}
```

What do you think @mubaidr ?

* I based myself on https://kulshekhar.github.io/ts-jest/user/config/

mubaidr:
Sound good enough! 

lets use `ts-jest/presets/js-with-ts` preset and avoid `babel` for tests only by extending `tsconfig.json` to `tsconfig.test.json`

eliamartani:
Changes made. Tested with the preset and it worked.

PR #648, opened 2020-10-24
Update es5 require to es6 import on test folder

eliamartani:
Following the same approach I did on #647

## Description
This code doesn't fix anything other than keeping codebase equal.

## Motivation and Context
It doesn't solve any issue, only keep things organized

## How Has This Been Tested?
The same way previously, first I ran `npm run lint` to remove some warnings and second `npm run build` to guarantee it is not adding any issues

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/__tests__/istanbul-linker-util.ts
+++ b/__tests__/istanbul-linker-util.ts
@@ -142,7 +142,12 @@ class Recurse {
   }
 }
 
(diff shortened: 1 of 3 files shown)


mubaidr:
https://github.com/BrainJS/brain.js/pull/647#issuecomment-716029891

PR #652, opened 2020-10-27
migrate examples/javascript/predict-tense.js to examples/typescript/predict-tense.ts

luisgregson:
Creates an equivalent typescript example for the existing javascript predict-tense example

## Description
Copied the examples/javascript/predict-tense.js file to examples/typescript/predict-tense.ts and fixed import format.

## Motivation and Context
This is part of a migration to typescript covered by issue [#551](https://github.com/BrainJS/brain.js/issues/551)

## How Has This Been Tested?
Ran the linter and compiled the file with the tsc binary in node_modules

N/A

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:

### DIFF
--- a/examples/typescript/predict-tense.ts
+++ b/examples/typescript/predict-tense.ts
@@ -1,0 +1,53 @@
+import { brain } from '../../src';
+
+// create configuration for training
+const config = {
+  iterations: 15000,
+  log: true,
+  logPeriod: 500,
+  layers: [10],
+};
+
+// create data which will be used for training
+const data = [
+  { input: 'I will write a book.', output: 'future' },
+  { input: 'He will be writing a book.', output: 'future' },
+  { input: 'They will have written the book.', output: 'future' },
+  { input: 'I will not cry.', output: 'future' },
+  { input: 'You will find happiness', output: 'future' },
+  { input: 'I am about to leave.', output: 'future' },
+  { input: 'Will you go out?', output: 'future' },
+  { input: 'I shall leave.', output: 'future' },
+  { input: 'We will rock you.', output: 'future' },
+  { input: 'I shall bring the laptop.', output: 'future' },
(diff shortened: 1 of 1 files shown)


examples/typescript/predict-tense.ts
```diff
@@ -0,0 +1,53 @@
...
+  { input: 'Am I wrong?', output: 'present' },
+  { input: 'Right this way.', output: 'present' },
+];
+
+// the thing we would test
+const test = 'Will you be my friend?';
+
+const network = new brain.recurrent.LSTM();
+network.train(data, config);
+const output = network.run(test);
```
mubaidr, line 52: we can cast result to `string`

mubaidr (changes requested):
Cast output to proper type i.e. `string` in our case

examples/typescript/predict-tense.ts
luisgregson: I'm not sure i understand. Can you give me an example of how you would like it to look?

The template literal automatically casts `output` to string, and i think it makes sense for the `output` variable to infer its type from the  `network.run` method?

I could give `output` the type of string (`const output: String = network.run(test);`) in order to cause an error if `run` is typed to return something else?

luisgregson: or i could make sure it is always a string regardless of what `network.run(test)` returns by doing this?
```
const output = String(network.run(test));
```

luisgregson:
> Cast output to proper type i.e. `string` in our case

I've added a question to the comment, as i wasn't sure what the change should be

examples/typescript/predict-tense.ts
mubaidr: Good point! The result will be typed once we have `rnn` network typed.

PR #677, opened 2020-12-28
Rnn time step ts

robertleeplummerjr:

## Description
This is mostly just conversion to typescript from javascript, that being said, because this is one of the more complex classes, there are some backwards compatibility issues.  Namely anything that had a line on it that said "backwards compatibility", because we can only maintain so much.  And second, dealing with objects is a bit tricky with typescript, though fully supported.

## Motivation and Context

## Types of changes

## Author's Checklist:

## Reviewer's Checklist:


robertleeplummerjr:
My goal here is to get this merged, and then start back on unit tests. Currently something like 80 fail, mostly edge cases.

